PR #2, opened 2020-06-23
Add /yt servlet

chenzizhao:
This is a skeleton file of the /yt servlet. It fetches first 20 comments of a specific video and prints them out to /yt page. Sentiment analysis, formatting the response into a json for the front end to parse and very likely changing url are needed. 

How to test: When you checkout the "yt-comment" branch, insert the `DEV_KEY`, run the test server and go to /yt, it should print out the first 20 comments of the sample video (without requiring any authorization).


chenzizhao:
Duplicate servlet with #7

PR #4, opened 2020-06-25
add cloudbuild files

nif33:
Add files to enable cloudbuild
(Cloudbuild seems to be a mandatory check in this repository)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+FROM alpine
+COPY src/main/webapp/index.html /
+CMD ["/src/main/webapp/index.html"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM alpine
+COPY src/main/webapp/index.html /
+CMD ["/src/main/webapp/index.html"]
```
nif33, line 3: I'm not sure if this is the right source for the Dockerfile... was very confused by the documentation but it seems to work and it is I guess the origin of the app?

PR #5, opened 2020-06-25
get captions for video requested by user

nif33:
adds an html form for user to input the video URL
gets captions for the video and parses it into JSON format

### DIFF
--- a/src/main/webapp/scripts/captions.js
+++ b/src/main/webapp/scripts/captions.js
@@ -1,0 +1,20 @@
+
+function getCaptions() {
+  videoId = document.getElementById("video-url").value.split('v=')[1];
+  fetch(`http://video.google.com/timedtext?lang=en&v=${videoId}`)
+    .then(response => response.text())
+    .then((responseText) => {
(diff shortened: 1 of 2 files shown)


src/main/webapp/scripts/captions.js
```diff
@@ -0,0 +1,20 @@
+
+function getCaptions() {
+  videoId = document.getElementById("video-url").value.split('v=')[1];
+  fetch(`http://video.google.com/timedtext?lang=en&v=${videoId}`)
+    .then(response => response.text())
+    .then((responseText) => {
+      const parser = new DOMParser();
```
chenzizhao, line 7: nit: maybe add a comment here to explain that this doesn't work with auto generated captions

nif33: I'll be adding a block later on to check whether the user input URL is a valid URL/video has captions so it'll be obvious in the code and to the user later

chenzizhao: 👌

PR #6, opened 2020-06-25
add comment toxicity checker with Perspective API

nif33:
creates a request object for the user inputted comment
gets comment toxicity analysis using perspective API
displays toxicity as a percentage to the user

### DIFF
--- a/src/main/webapp/scripts/perspective.js
+++ b/src/main/webapp/scripts/perspective.js
@@ -1,0 +1,21 @@
+function getToxicity() {
+  API_KEY = '';
+  const commentText = document.getElementById("comment-text").value;
+  const commentObject = {
+    'comment': {'text': commentText},
+    'requestedAttributes': {'TOXICITY': {}},
+    'languages': ['en']
+  };
+  const json = JSON.stringify(commentObject);
+  const request = new Request(
+    `https://commentanalyzer.googleapis.com/v1alpha1/comments:analyze?key=${API_KEY}`,
+    {method: 'POST', body: json}
+  );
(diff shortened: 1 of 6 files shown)


src/main/webapp/scripts/perspective.js
```diff
@@ -0,0 +1,22 @@
...
+  };
+  const json = JSON.stringify(commentObject);
+  const request = new Request(
+    `https://commentanalyzer.googleapis.com/v1alpha1/comments:analyze?key=${API_KEY}`,
+    {method: 'POST', body: json}
+  );
+  fetch(request)
+    .then(response => response.text())
+    .then((responseText) => {
+      const data = JSON.parse(responseText);
```
chenzizhao, line 17: not 100% sure but this should also work? 
```
fetch(request)
.then(response => response.json())
.then(data => {
const score = data.attributesScores ...
})
```

src/main/webapp/scripts/perspective.js
```diff
@@ -0,0 +1,22 @@
...
+      const score = data.attributeScores.TOXICITY.summaryScore.value;
+      containerDOM = document.getElementById('toxicity-container');
```
chenzizhao, line 19: nit: use `const` by default

https://google.github.io/styleguide/jsguide.html#features-use-const-and-let

src/main/webapp/scripts/perspective.js
nif33: I tried it earlier but I must've done something wrong because it works now haha, changed!

PR #7, opened 2020-06-25
Get video comments and calculate sentiment.

kayefung:
Creates a String of comments and calculates the comments' collective sentiment score, after which the comments and the score can be fetched from /yt.

All the comments are being concatenated into one String, because the NLP API can only analyze the sentiment of a single String or data that's in gCloud storage, which we could implement later.


src/main/java/com/google/sps/data/Video.java
```diff
@@ -0,0 +1,49 @@
...
+        this.comments = comments;
+        this.commentScore = calculateSentimentScore(convertListToString(comments));
+    }
+
+    // Concatenates all Strings of comments into one String.
+    private String convertListToString(List<String> list) {
+        String str = "";
+        for (String text : list) {
+            str = str.concat(text + ". ");
+        }
```
chenzizhao, line 25: How about using String.join()?

https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-

src/main/webapp/scripts/comments.js
```diff
@@ -0,0 +1,36 @@
...
+        }).then(function(response) {
+            let commentsList = new Array();
+            for (commentThread of response.result.items) {
+                let commentContent = commentThread.snippet.topLevelComment.snippet.textDisplay;
+                commentsList.push(commentContent);
+            }
+
+            fetch('/yt', {
+                method: 'POST',
+                body: JSON.stringify(commentsList),
```
chenzizhao, line 24: glad it works out!

src/main/java/com/google/sps/servlets/YouTubeServlet.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String commentsStr = request.getReader().lines().collect(Collectors.joining());
+        Gson gson = new Gson();
+        String[] commentsArr = gson.fromJson(commentsStr, String[].class);
+        List<String> commentsList = Arrays.asList(commentsArr);
+        video = new Video(commentsList);
+
+        response.sendRedirect("/index.html");
```
chenzizhao, line 37: Since we are fetching from the front end, I assume redirect isn't needed?

src/main/java/com/google/sps/servlets/YouTubeServlet.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.stream.Collectors;
+import com.google.sps.data.Video;
+
+@WebServlet("/yt")
+public class YouTubeServlet extends HttpServlet {
+
+    private Video video;
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
chenzizhao, line 21: I am not sure if this is intended -- where is "/yt" GET called?

src/main/webapp/scripts/comments.js
```diff
@@ -0,0 +1,36 @@
+const key = 'DEV_KEY_HERE'; //<- change as needed
+const ytDiscovery = 'https://www.googleapis.com/discovery/v1/apis/youtube/v3/rest';
+const languageDiscovery = 'https://language.googleapis.com/$discovery/rest?version=v1beta2';
+
+function loadClient() {
+    gapi.client.init({
+        'apiKey': key,
+        'discoveryDocs': [ytDiscovery, languageDiscovery],
+    }).then(function() {
```
nif33, line 9: I believe there is a way to have the getting comments part separate from the client initiation. I think it'd be good to separate them because the function name loadClient doesn't make it clear that its loading the Client then also getting comments

src/main/webapp/scripts/comments.js
kayefung: Thanks for helping me out with it guys:)

src/main/java/com/google/sps/servlets/YouTubeServlet.java
kayefung: True! Done.

src/main/java/com/google/sps/data/Video.java
kayefung: Nice one! Done.

src/main/java/com/google/sps/servlets/YouTubeServlet.java
kayefung: Yeah so far I haven't added this part yet. Probably will be added when we put the comments and captions features together.

src/main/webapp/scripts/comments.js
```diff
@@ -0,0 +1,36 @@
...
+            fetch('/yt', {
+                method: 'POST',
+                body: JSON.stringify(commentsList),
+                headers: {
+                    'Content-type': 'application/json' 
+                }
+            });
+
+        }, function(err) {
+            console.error("Execute error", err);
```
nif33, line 31: Can you add to the error message to make it more specific for this error?

src/main/java/com/google/sps/servlets/YouTubeServlet.java
chenzizhao: I see I see

src/main/webapp/scripts/comments.js
nif33: Tested some things, you don't need to put the ``` return gapi.client.youtube.commentThreads.list({``` in the then statement at all so you could do something like
```
loadClient() {
}
getComments(){
  loadClient()
  // do comment things
}
```

src/main/webapp/scripts/comments.js
```diff
@@ -1,36 +1,52 @@
...
 
-        }, function(err) {
-            console.error("Execute error", err);
-        });
+function getComments(callback) {
+  let commentsList = new Array();
+  
+  try {
+    const request = gapi.client.youtube.commentThreads.list({
+      'part': ['snippet,replies'],
```
chenzizhao, line 21: Probably we can get rid of "reply" because we are only reading the top level comment?

src/main/java/com/google/sps/data/Video.java
```diff
@@ -0,0 +1,39 @@
...
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.Sentiment;
+
+public class Video {
+
+  private final List<String> comments;
+  private final float commentScore;
+
+  public Video(List<String> comments/*, List<String> captions*/) throws IOException {
+    // Can add captions functionality later on.
```
juanalonsocruz, line 15: Nit: Remove comments of features that don't exist yet.

src/main/java/com/google/sps/data/Video.java
```diff
@@ -0,0 +1,39 @@
...
+  private float calculateSentimentScore(String text) throws IOException {
+    Document doc =
+        Document.newBuilder().setContent(text).setType(Document.Type.PLAIN_TEXT).build();
+    LanguageServiceClient languageService = LanguageServiceClient.create();
+    Sentiment sentiment = languageService.analyzeSentiment(doc).getDocumentSentiment();
+    float score = sentiment.getScore();
+    languageService.close();
+
+    return score;
+  }
```
juanalonsocruz, line 30: Classes under data directories tend to be used just to represent data in objects without having other methods that act on  the object. Maybe move this as a helper method to servlet file. Calculate elsewhere and pass the answer to this constructor.

src/main/java/com/google/sps/data/Video.java
```diff
@@ -0,0 +1,39 @@
+package com.google.sps.data;
+
+import java.io.IOException;
+import java.util.List;
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.Sentiment;
+
+public class Video {
```
juanalonsocruz, line 9: `Video` makes it seem like this class represent what a video. It is more representing of something like `VideoCommentsAnalysis` or something along those lines.

src/main/java/com/google/sps/servlets/YouTubeServlet.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Arrays;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.gson.Gson;
+import java.util.stream.Collectors;
+import com.google.sps.data.Video;
+
+@WebServlet("/yt")
```
juanalonsocruz, line 15: Nit: Servlet method should be representative of what it does (just like method names).

src/main/java/com/google/sps/servlets/YouTubeServlet.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String commentsStr = request.getReader().lines().collect(Collectors.joining());
+    Gson gson = new Gson();
+    String[] commentsArr = gson.fromJson(commentsStr, String[].class);
+    List<String> commentsList = Arrays.asList(commentsArr);
+    video = new Video(commentsList);
+  }
```
juanalonsocruz, line 36: There is some odd behavior happening in this servlet.

Since video is a private property of the servlet class, you could have a case where user A makes the POST request which populates the video property and then user B makes a GET request which would get the data that user A requested.

GET methods are used to get some kind of information based on the request. POST methods are used to store data that is sent in the request somewhere like in a database. Since you are not getting data from the request that will store somewhere, you don't need a POST method. In this case, the requester gives you some data and is requesting the sentiment score. This seems more like GET.

What I would expect here is have a GET method that takes in the comments from the request, calls the sentiment API, puts the score into the data object, JSONify it, and then send it back.

This change would require client side changes. But like I mention in another comment dont think about the client side code until the back-end is developed and have them as separate PRs. If you are blocking a client side developer from coding, just tell them what the API (emphasis on `interface`) is. They shouldn't care about how back end does the processing.

src/main/java/com/google/sps/data/Video.java
juanalonsocruz, line 30: You made a comment about how the API only takes one string to analyze. We had discuss that merging all comments might give you a different sentiment because it might be reading it as one blob of text from a singular person than if you gave call the API once per comment and then say averaged the score. To make multiple calls to the sentiment API efficiently, you would need to call the API asynchronously using Java Threads or Futures. I would recommend Java 8 Futures. Since threads and Future might be a lot to add for this PR, I would accept the merging comments implementation for now but expect this to be improved in the next PR after researching on how to use Futures.

src/main/webapp/scripts/comments.js
```diff
@@ -0,0 +1,53 @@
+const key = 'DEV_KEY_HERE'; //<- change as needed
```
juanalonsocruz, line 1: I recommend the client side code to be in a separate PR that needs to be strictly merged after the back-end code is merged. Recommending this because, we will enforce any PR to also have unit tests and that will be a lot of line changes in one PR for reviewers to review.

src/main/java/com/google/sps/servlets/YouTubeServlet.java
```diff
@@ -0,0 +1,37 @@
+package com.google.sps.servlets;
```
juanalonsocruz, line 1: Missing Java unit tests. Please ask if you are not sure on what to actually test. Making unit tests a requirement before merging any code.

src/main/webapp/scripts/comments.js
juanalonsocruz, line 1: Add JS unit tests.

kayefung: Should I be removing the client side code from this PR?

juanalonsocruz: Your choice. But it could potentially lead to a big PR with both java and js unit tests.

kayefung: Okay, I'm going to remove the comments.js file from this PR and reintroduce it in a different PR (or call comments from the backend instead).

kayefung: Ack

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -0,0 +1,70 @@
+package com.google.sps;
```
juanalonsocruz, line 1: Comment for the whole test:

You want to test behavior of your servlet and not the helper methods. I recommend reading ToTTs `Test Behaviors, Not Methods` and `Test Behavior, Not Implementation`.

Basically mock your http requests such that it handles your edge cases (even if they are handled just in helper methods). You want test how a client uses your servlet directly. Nobody is calling your helper methods directly.

How to test HTTP servlet stuff (it even includes an example on how to test Writers. Maybe you could do something similar with Readers since you are using it in your implementation).
https://stackoverflow.com/questions/5434419/how-to-test-my-servlet-using-junit

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -23,48 +28,71 @@
...
   @Test
-  public void jsonArrayConvertsToPeriodSeparatedString() throws IOException {
+  public void jsonArrayPrintsVideoAnalysisJsonObject() throws IOException {
     String json = "[\"a\",\"b\",\"c\"]"; // "["a","b","c"]"
     when(requestMock.getReader()).thenReturn(new BufferedReader(new StringReader(json)));
     
-    String actual = sentimentServlet.requestToString(requestMock);
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(responseMock.getWriter()).thenReturn(writer);
```
juanalonsocruz, line 47: (1) Is this needed? Does something like `responseMock.getWriter().toString()` work?
(2) If it is needed, it seems like you need this for all your test cases so you can verify the response. You can have a method with the `@Before` attribute that automatically runs before each test. https://www.baeldung.com/junit-before-beforeclass-beforeeach-beforeall

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -23,48 +28,71 @@
...
+
+    new SentimentServlet().doGet(requestMock, responseMock);
```
juanalonsocruz, line 49: You already declared a SentimentServlet object above. It is inside the variable `sentimentServlet`

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -23,48 +28,71 @@
...
 
-    Assert.assertEquals("a. b. c", actual);
+    String pattern = "\\{\"commentScore\":[-+]?[0-9]*\\.?[0-9]+\\}\n";
```
juanalonsocruz, line 51: Is this unit test actually getting a response from the sentiment API which is why you are using regex to catch any score that it returns?

Unit tests should only test the direct file they are testing. You are technically doing this by doing the regex and not caring about the actual score but if its actually making a service call then your unit test might fail if the service goes down. Instead, you should mock with a hardcoded value what the sentiment API returns and in your unit test you can verify that the value the API returned was correctly translated into your servlet response.

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -23,34 +23,45 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String comments = requestToString(request);
```
juanalonsocruz, line 25: Instead of passing null through the whole servlet, having all methods need to handle null, and eventually returning an empty JSON object (potentially having the client confused why it returned an empty object), you could catch the syntax error, and return an HTTP error explaining to the client that something went wrong (this case JSON syntax error).

This way the client knows what went wrong and you don't have to handle null values everywhere.

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -23,48 +28,71 @@
...
-   * If the Json String is an empty array, an empty String should be returned. 
-   * Ex. "[]" -> ""
+   * If the Json String is not in the form of an array, a JsonSyntaxException is caught and null is
+   * returned. Gson excludes null values, therefore resulting in an empty Json object.
+   * Ex. "foo" -> "{}"
    */
   @Test
-  public void emptyJsonArrayConvertsToEmptyString() throws IOException {
-    String json = "[]";
+  public void notArrayPrintsEmptyJsonObject() throws IOException {
```
juanalonsocruz, line 62: change this test after addressing the comment above to assert if an error was return in the HTTP response.

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -23,34 +23,45 @@
...
+   */
+  private String requestToString(HttpServletRequest request) throws IOException {
     Gson gson = new Gson();
     
     String textStr = request.getReader().lines().collect(Collectors.joining());
-    List<String> textList;
+
+    if (textStr.equals("[]")) {
+      return null;
+    }
```
juanalonsocruz, line 46: How about letting gson convert this into an empty list which then the string join will return an empty string? Seems like the intended behavior of a method called `requestToString`. Then your main method could handle this. I recommend not even moving forward with the rest of the servlet with an empty string and returning an HTTP bad request error saying that the response must be a non-empty array.

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -23,48 +28,71 @@
...
+    Assert.assertTrue(stringWriter.toString().matches(pattern));
   }
 
   /**
-   * If the Json String is not in the form of an array, a JsonSyntaxException is caught and an
-   * empty String is returned. 
-   * Ex. "foo" -> ""
+   * If the Json String is an empty array, a JsonSyntaxException is caught and null is returned.
+   * Gson excludes null values, therefore resulting in an empty Json object.
+   * Ex. "[]" -> "{}"
```
juanalonsocruz, line 80: Change this test after addressing the comment above to verify that it returns the specific HTTP error.

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -23,34 +23,45 @@
...
+   * Calculates sentiment score of text. The score is from -1 (negative) to +1 (positive).
+   */ 
+  private Float calculateSentimentScore(String text) throws IOException {
+    if (text == null) {
+      return null;
+    }
+
     Document doc =
         Document.newBuilder().setContent(text).setType(Document.Type.PLAIN_TEXT).build();
     LanguageServiceClient languageService = LanguageServiceClient.create();
```
juanalonsocruz, line 67: Can something go wrong in the language service? If so, how to they tell you that something went wrong? How should handle this? The front end might be interested to know if something in stack failed. Would recommend returning an HTTP error explaining that the language service failed for a reason not related to the request in this servlet.

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -0,0 +1,92 @@
...
+
+    try {
+      String comments = requestToString(request);
+      if (comments.equals("")) {
+        response.sendError(HttpServletResponse.SC_BAD_REQUEST,
+            "No comments to analyze. Request must be a non-empty array");
+        isValid = false;
+      }
+      commentsScore = calculateSentimentScore(comments);
+    } catch (JsonSyntaxException e) {
```
juanalonsocruz, line 42: I also recommend you log any exceptions that are sent to the client in the Java side with System.err.println(e.getMessage()). This way you can see the real details of the exceptions in the back-end. There are other ways to log in Java (i.e. Logger) but that could be done in a future PR or just something to consider for your future internships/career.

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -0,0 +1,113 @@
...
+   * error is sent.
+   */
+  @Test
+  public void notArraySendsHttpResponseError() throws IOException {
+    String json = "foo";
+    when(requestMock.getReader()).thenReturn(new BufferedReader(new StringReader(json)));
+    
+    sentimentServlet.doGet(requestMock, responseSpy);
+
+    verify(responseSpy, times(1)).sendError(eq(HttpServletResponse.SC_BAD_REQUEST), anyString());
```
juanalonsocruz, line 82: Nit: I believe you don't need the times(1). Its that by default.

juanalonsocruz, line 82: Replace anyString() with the actual string sent or or containsString. You send two different BAD_REQUEST. You want to make sure you are sending a specific one.

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -0,0 +1,113 @@
...
+   * If the Json String is an empty array (ex. no comments), a 400 error is sent.
+   */
+  @Test
+  public void emptyArraySendsHttpResponseError() throws IOException {
+    String json = "[]";
+    when(requestMock.getReader()).thenReturn(new BufferedReader(new StringReader(json)));
+    
+    sentimentServlet.doGet(requestMock, responseSpy);
+
+    verify(responseSpy, times(1)).sendError(eq(HttpServletResponse.SC_BAD_REQUEST), anyString());
```
juanalonsocruz, line 95: Replace anyString with the actual string or contains string.

juanalonsocruz, line 95: Remove times(1)

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * If the Natural Language API fails and throws an ApiException, a 500 error is sent.
+   */
+  @Test
+  public void languageServiceFailureSendsHttpResponseError() throws IOException {
+    String json = "[\"a\",\"b\",\"c\"]";
```
juanalonsocruz, line 103: Nit: This is irrelevant data to this test because you are mainly testing how you handle the service failure. I would create a constant json string pass it directly into StringReader(...).

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -0,0 +1,113 @@
...
+    when(requestMock.getReader()).thenReturn(new BufferedReader(new StringReader(json)));
+    when(languageServiceMock.analyzeSentiment(any(Document.class))).thenThrow(ApiException.class);
+
+    sentimentServlet.doGet(requestMock, responseSpy);
+
+    verify(responseSpy, times(1))
```
juanalonsocruz, line 109: Remove times(1)

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -0,0 +1,113 @@
...
+        .sendError(eq(HttpServletResponse.SC_INTERNAL_SERVER_ERROR), anyString());
```
juanalonsocruz, line 110: Replace anyString with actual string.

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * A Json String in the form of an array should be converted to a String of period separated
+   * comments. A sentiment score should be calculated and the score should be printed in a Json
+   * object. 
+   * Ex. "[foo, bar]" -> "{"commentScore":0.05}" 
+   */
+  @Test
+  public void jsonArrayPrintsVideoAnalysisJsonObject() throws IOException {
```
juanalonsocruz, line 58: Recommend adding another test with an array of 1 element. Just to make sure it works with the simple case as well. Remember to change the name of tests as well. `jsonSingleElementArray...` and `jsonMultiElementArray...`

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -0,0 +1,92 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Float commentsScore = null;
+    boolean isValid = true;
+
+    try {
+      String comments = requestToString(request);
+      if (comments.equals("")) {
+        response.sendError(HttpServletResponse.SC_BAD_REQUEST,
+            "No comments to analyze. Request must be a non-empty array");
+        isValid = false;
```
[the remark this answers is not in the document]
kayefung: doGet is a void method (and also overrides a method of its superclass so I can't change it), so I used isValid to avoid running the rest of the method if there's an error.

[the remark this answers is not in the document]
kayefung: nvm, Zoe helped me out with this!

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+    
+  }
+
+  /**
+   * Converts Json String of a request (that is in the form of an array) to a String, separated by
+   * periods. If the Json cannot be converted to an array or is an empty array, null is returned. 
+   * Ex. "["a","b","c"]" --> "a. b. c"
+   */
+  private String requestToString(HttpServletRequest request) throws IOException {
```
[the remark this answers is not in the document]
kayefung: Ack, will do in an upcoming PR!

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -28,34 +28,43 @@ public SentimentServlet() throws IOException {
...
+          "Request to /sentiment must be an array of Strings.");
+      return;
+    } catch (IOException e) {
+      System.err.println(e.getMessage());
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+          "Converting request to String failed.");
+      return;
+    }
+
+    if (comments.equals("")) {
```
juanalonsocruz, line 48: Nit: `comments.isEmpty()`

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -79,7 +100,8 @@ public void notArraySendsHttpResponseError() throws IOException {
     
     sentimentServlet.doGet(requestMock, responseSpy);
 
-    verify(responseSpy, times(1)).sendError(eq(HttpServletResponse.SC_BAD_REQUEST), anyString());
+    verify(responseSpy).sendError(HttpServletResponse.SC_BAD_REQUEST, 
+        "Request to /sentiment must be an array of Strings.");
```
juanalonsocruz, line 104: Nit: Make these strings into constant string variables

PR #8, opened 2020-07-02
Fetch caption in backend

chenzizhao:
To avoid transferring long strings from the front end to the back end for sentiment analysis, the `Caption` class and the `CaptionService` class are implemented to fetch English captions (if exists) given a valid video id. 

`CaptionService` class can open a `HttpUrlConnection` and returns an input stream, then the `Caption` class processes the input stream and returns a caption string, ready for sentiment analysis. 

The endpoint is the `Caption.getCaptionFromId()` function: this can be called in a servlet which is exposed to the front end. 

The `Caption` class is unit-tested by mocking its dependency `CaptionService`. `CaptionService` class is not unit-tested because it opens actual http connections (I did manually tested the whole flow and it worked). Any suggestion on how to test the `CaptionService` class is appreciated.

### DIFF
--- a/src/main/java/com/google/sps/servlets/Caption.java
+++ b/src/main/java/com/google/sps/servlets/Caption.java
@@ -10,7 +10,7 @@ import org.xml.sax.SAXException;
 
 public class Caption {
   /*
-   * Usage: String caption = new Caption(new CaptionService()).getCaptionFromID("abcdefg");
+   * Usage: String caption = new Caption().getCaptionFromID("abcdefg");
    * 
    * Return null if not found.
    */
--- a/src/main/java/com/google/sps/servlets/Caption.java
+++ b/src/main/java/com/google/sps/servlets/Caption.java
@@ -1,0 +1,46 @@
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import java.io.InputStream;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+public class Caption {
+  /*
+   * Usage: String caption = new Caption(new CaptionService()).getCaptionFromID("abcdefg");
+   * 
+   * Return null if not found.
+   */
+
+  private CaptionService captionService = new CaptionService();
+  private DocumentBuilderFactory docBuilderFactory = DocumentBuilderFactory.newInstance();
+
+  public String getCaptionFromId(String videoId) {
+    // BCP47 language tags
+    final String[] languages = {"en", "en-US", "en-GB", "en-AU", "en-IE", "en-ZA"};
+    for (String language : languages) {
+      try {
+        return parseXmlFromStream(captionService.fetchStream(videoId, language));
+      } catch (SAXException | ParserConfigurationException | IOException
+          | IllegalArgumentException e) {
+        continue;
+      }
+    }
+    return null;
+  }
+
+  private String parseXmlFromStream(InputStream inputStream)
+      throws SAXException, ParserConfigurationException, IOException, IllegalArgumentException {
+    Document doc = docBuilderFactory.newDocumentBuilder().parse(inputStream);
+    NodeList textDom = doc.getElementsByTagName("text");
+    StringBuilder caption = new StringBuilder();
+    for (int i = 0; i < textDom.getLength(); i++) {
+      caption.append(textDom.item(i).getFirstChild().getNodeValue());
+      caption.append(" ");
+    }
+    return caption.toString().trim();
+  }
+}
--- a/src/main/java/com/google/sps/servlets/CaptionService.java
+++ b/src/main/java/com/google/sps/servlets/CaptionService.java
@@ -2,14 +2,12 @@ package com.google.sps.servlets;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.MalformedURLException;
 import java.net.URL;
 
 public class CaptionService {
-  public InputStream fetchStream(String videoId, String language)
-      throws MalformedURLException, IOException {
+  public InputStream fetchStream(String videoId, String language) throws IOException {
     if (videoId == null) {
-      throw new MalformedURLException();
+      throw new IllegalArgumentException("Video ID must be a non-null string.");
     }
     return new URL(
         String.format("http://video.google.com/timedtext?lang=%s&v=%s", language, videoId))
--- a/src/main/java/com/google/sps/servlets/CaptionService.java
+++ b/src/main/java/com/google/sps/servlets/CaptionService.java
@@ -1,0 +1,18 @@
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class CaptionService {
+  public InputStream fetchStream(String videoId, String language)
+      throws MalformedURLException, IOException {
+    if (videoId == null) {
+      throw new MalformedURLException();
+    }
+    return new URL(
+        String.format("http://video.google.com/timedtext?lang=%s&v=%s", language, videoId))
+            .openStream();
+  }
+}
--- a/src/test/java/com/google/sps/servlets/CaptionTest.java
+++ b/src/test/java/com/google/sps/servlets/CaptionTest.java
@@ -1,0 +1,125 @@
+package com.google.sps.servlets;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.contains;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.AdditionalMatchers.not;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+@RunWith(JUnit4.class)
+public final class CaptionTest {
+  @Mock
+  private CaptionService captionServiceMock;
+
+  @InjectMocks
+  private Caption caption;
+
+  final private String PROPER_XML = "<?xml version=\"1.0\" encoding=\"utf-8\" ?><transcript>"
+      + "<text start=\"4.029\" dur=\"5.401\">line1</text>"
+      + "<text start=\"9.43\" dur=\"5.06\">line2</text>"
+      + "<text start=\"14.49\" dur=\"5.03\">line3</text>" + "</transcript>";
+
+  final private String PROPER_XML_CAPTION = "line1 line2 line3";
+
+  final private String BAD_XML = "<?xml version=\"1.0\" encoding=\"utf-8\" ?><transcript>"
+      + "<text start=\"4.029\" dur=\"5.401\">line1</text>";
+
+  final private String VIDEO_ID = "sampleId";
+
+  final private String LANG_EN = "en";
+
+  final private String LANG_EN_GB = "en-GB";
+
+  @Before
+  public void init() {
+    caption = new Caption();
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void properXmlReturnsConcatenatedString() throws IOException {
+    InputStream stream = new ByteArrayInputStream(PROPER_XML.getBytes());
+    when(captionServiceMock.fetchStream(eq(VIDEO_ID), contains(LANG_EN))).thenReturn(stream);
+
+    String actual = caption.getCaptionFromId(VIDEO_ID);
+
+    assertEquals(PROPER_XML_CAPTION, actual);
+  }
+
+  @Test
+  public void badXmlReturnsNullIgnoringSAXException() throws IOException {
+    InputStream stream = new ByteArrayInputStream(BAD_XML.getBytes());
+    when(captionServiceMock.fetchStream(eq(VIDEO_ID), contains(LANG_EN))).thenReturn(stream);
+
+    String actual = caption.getCaptionFromId(VIDEO_ID);
+
+    assertNull(actual);
+  }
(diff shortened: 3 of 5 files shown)


src/main/java/com/google/sps/servlets/CaptionService.java
```diff
@@ -0,0 +1,18 @@
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class CaptionService {
```
kayefung, line 8: Is there a reason for having Caption and CaptionService as separate classes? Just wondering!

src/test/java/com/google/sps/servlets/CaptionTest.java
```diff
@@ -0,0 +1,109 @@
...
+    InputStream stream = new ByteArrayInputStream(PROPER_XML.getBytes());
+    when(captionServiceMock.fetchStream(eq(VIDEO_ID), contains(LANG_EN))).thenReturn(stream);
+
+    String actual = new Caption(captionServiceMock).getCaptionFromId(VIDEO_ID);
+    String expected = PROPER_XML_CAPTION;
+    Assert.assertEquals(expected, actual);
+  }
+
+  @Test
+  public void badXml() throws IOException {
```
kayefung, line 61: Use more descriptive names for your tests so it's obvious what is being tested and how. ToTT #339, Writing Descriptive Test Names, goes more into detail about it.

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -0,0 +1,49 @@
...
+          | IllegalArgumentException e) {
+        continue;
+      }
+    }
+    return null;
+  }
+
+  private String parseXmlFromStream(InputStream inputStream)
+      throws SAXException, ParserConfigurationException, IOException, IllegalArgumentException {
+    Document doc = DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(inputStream);
```
nif33, line 40: NIT: indentation is off here

src/main/java/com/google/sps/servlets/CaptionService.java
chenzizhao: I can relate it is annoying to have two classes instead of one 😅 The reason is to avoid actually making Internet connections/API calls when testing.

src/test/java/com/google/sps/servlets/CaptionTest.java
chenzizhao: Done. I like how you named your tests!

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -0,0 +1,53 @@
...
+   * Usage: String caption = new Caption().getCaptionFromID("abcdefg");
+   * 
+   * Return null if not found.
+   */
+
+  public CaptionService captionService;
+
+  public Caption() {
+    captionService = new CaptionService();
+  }
```
juanalonsocruz, line 22: I'm not sure what is preferred as good practice but I have it a hunch that it might be preferred to do

```
public CaptionService captionService = new CaptionService;
```

and not have a custom constructor (If you don't have a constructor, there is a default constructor behavior).


I would use constructor setters if the non-test caller cares about setting specific members or if there are different overloaded constructors that behave differently.

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public void setCaptionService(CaptionService captionService) {
+    this.captionService = captionService;
+  }
```
juanalonsocruz, line 26: Also not sure about good practice here but I have another hunch lol.

Being able to reset a dependency  after the object has been created seems like it could cause issues in the future. For example, another developer resets the dependency cause it works for their use case but breaks someone else who using the same object instance (of course these behavior should be tested on any file that uses this class but can't guarantee that). So instead, we would like to gear users to create their own instance for their use cases instead of resetting. This can be enforced by making CaptionService a private member and have it be passed in the constructor.

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private String parseXmlFromStream(InputStream inputStream)
+      throws SAXException, ParserConfigurationException, IOException, IllegalArgumentException {
+    Document doc = DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(inputStream);
+    NodeList textDom = doc.getElementsByTagName("text");
+    String caption = "";
+    int lenTextDom = textDom.getLength();
+    for (int i = 0; i < lenTextDom; i++) {
+      caption = caption + " " + textDom.item(i).getFirstChild().getNodeValue();
+    }
```
juanalonsocruz, line 50: Algorithmic complexity here is O(|caption|^2). Java Strings are immutable and when you concatenate two strings, it is actually building a new string from scratch taking characters from both input strings. Instead of concatenating a new strings you want to append new characters into an existing mutable string object. StringBuilder is a good candidate. 

FYI: This is a common mistake in Java/Python interviews. C++ concatenating with "+" creates a new string so it has the same issue but using "+=" actually appends to the string efficiently because C++ strings are mutable.

src/main/java/com/google/sps/servlets/Caption.java
juanalonsocruz: I see you learned about InjectMock. No need for this setter anymore lol.

src/main/java/com/google/sps/servlets/CaptionService.java
juanalonsocruz, line 8: You had a comment about testing CaptionService. Hmm... maybe you don't need CaptionService at all if you can somehow mock the openStream method for URL class to return an stream that contains XML content you would like to test on. What do you think?

src/main/java/com/google/sps/servlets/CaptionService.java
```diff
@@ -0,0 +1,18 @@
...
+  public InputStream fetchStream(String videoId, String language)
+      throws MalformedURLException, IOException {
+    if (videoId == null) {
+      throw new MalformedURLException();
```
juanalonsocruz, line 12: This should be `IllegalArgumentException` videoId is not a URL. Also should have exception message explaining what went wrong such as `videoId must be a non-null String value`.

src/test/java/com/google/sps/servlets/CaptionTest.java
```diff
@@ -0,0 +1,116 @@
...
+      + "<text start=\"4.029\" dur=\"5.401\">line1</text>"
+      + "<text start=\"9.43\" dur=\"5.06\">line2</text>"
+      + "<text start=\"14.49\" dur=\"5.03\">line3</text>" + "</transcript>";
+
+  final private String PROPER_XML_CAPTION = "line1 line2 line3";
+
+  final private String BAD_XML = "<?xml version=\"1.0\" encoding=\"utf-8\" ?><transcript>"
+      + "<text start=\"4.029\" dur=\"5.401\">line1</text>";
+
+  final private String FAILED_CAPTION = null;
```
juanalonsocruz, line 39: null is not a String and it is not easy when reading tests that you are checking for null instead of a string that represents a failed caption. There should be an assertNull method in JUnit. With this you don't need this variable.

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -0,0 +1,53 @@
...
+  public String getCaptionFromId(String videoId) {
+    // BCP47 language tags
+    final String[] languages = {"en", "en-US", "en-GB", "en-AU", "en-IE", "en-ZA"};
+    for (String language : languages) {
+      try {
+        return parseXmlFromStream(captionService.fetchStream(videoId, language));
+      } catch (SAXException | ParserConfigurationException | IOException
+          | IllegalArgumentException e) {
+        continue;
+      }
```
juanalonsocruz, line 37: Since you are explicitly catching these specific exceptions, then you need to test each one of them explicitly. You would need to mock parseXmlFromStream or any of the methods inside that method that actually throw those errors to actually throw them and then test getCaptionFromId returns null. I see you doing this already for IOException. So keep it up with the others.

Again, since you an explicit list of exceptions to catch, then maybe you should throw a different one in a test and verify that getCaptionFromId eventually throws a different one?

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+    return null;
+  }
+
+  private String parseXmlFromStream(InputStream inputStream)
+      throws SAXException, ParserConfigurationException, IOException, IllegalArgumentException {
```
juanalonsocruz, line 43: Question: Is explicitly saying that these methods could throw these exceptions needed if this method is not the one creating the exception (only underlying methods are doing so). I would think you didn't need to do it because then every method in a chain of calls would need them and all methods in existence would need them lol.

src/test/java/com/google/sps/servlets/CaptionTest.java
```diff
@@ -0,0 +1,116 @@
...
+  public void init() {
+    caption = new Caption();
+    MockitoAnnotations.initMocks(this);
+  }
+
+  private InputStream streamOn(String xml) {
+    return new ByteArrayInputStream(xml.getBytes());
+  }
+
+  @Test
```
juanalonsocruz, line 57: Is the throws IOException needed. Seems smelly if all unit tests have it.

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Done. I wasn't aware of the default constructor, thank you for bringing this up!

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Done. Field-based injection is much cleaner than the previous ones.

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Done. Good catch! I guess this is a pitfall that I should be very careful about from now on

src/main/java/com/google/sps/servlets/CaptionService.java
chenzizhao: That was my first idea too! unfortunately, Mockito can't mock final objects such as a URL. VSE and other people suggested using a wrapper class like this

src/test/java/com/google/sps/servlets/CaptionTest.java
chenzizhao: The tests can't build if I remove it from the function signature. Is this not the case in other testing packages?

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Thats a good counterexample. Can I generalize them as one Exception, since they will be ignored anyway in `getCaptionFromId()`?

src/main/java/com/google/sps/servlets/CaptionService.java
chenzizhao: Done. Thanks for pointing it out!

src/test/java/com/google/sps/servlets/CaptionTest.java
chenzizhao: Done.

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: SAXException, ParseConfigurationException, and IllegalArgumentException are thrown by the private function `parseXmlFromStream` and handled(ignored) in the public function `getCaptionFromId`. I am not sure if I should test a private function and mock its behavior of throwing exceptions?

juanalonsocruz: Yup you want to test that your code ignores these exceptions. Is it possible to mock the methods thats throw these exceptions  and throw them rather than mocking the private method?

src/main/java/com/google/sps/servlets/Caption.java
juanalonsocruz: I guess you do need them. I guess keep as it is now. There are two types of exceptions as explained below.
https://stackoverflow.com/questions/11589302/why-is-throws-exception-necessary-when-calling-a-function

Now I know why you don't need to say that you throw exceptions like `IllegalArgumentException`

I don't recommend you change it to Exception because it makes it vague to callers. Also, this method doesn't know the caller wants to ignore them.

src/main/java/com/google/sps/servlets/CaptionService.java
juanalonsocruz: Looks like you can in the version of Mockito you are using. Not sure if even worth the effort anymore.
https://stackoverflow.com/questions/14292863/how-to-mock-a-final-class-with-mockito

src/test/java/com/google/sps/servlets/CaptionTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @Before
+  public void init() {
+    caption = new Caption();
+    MockitoAnnotations.initMocks(this);
+  }
+
+  private InputStream streamOn(String xml) {
+    return new ByteArrayInputStream(xml.getBytes());
+  }
```
juanalonsocruz, line 54: This is a simple one liner. Its okay to have not have a helper method for this.

src/test/java/com/google/sps/servlets/CaptionTest.java
juanalonsocruz: Which method in the unit test can throw IOException? Is it the ByteArrayInputStream constructor? Or is it something else? Trying to see if there is a way to clean it off.

chenzizhao: It is the captionServiceMock.fetchStream() call.

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: okie I will keep it as it is. Thanks!

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Good idea. I mocked the DocumentBuilderFactory that throws the execeptions and it worked!

src/main/java/com/google/sps/servlets/CaptionService.java
chenzizhao: Lol I wonder why I didn't find it but thank you! I have included the plugin I am sure it will be useful for testing APIs, because some objects returned by the YouTube API (i.e. `CommentThread`) are final objects too.

src/test/java/com/google/sps/servlets/CaptionTest.java
chenzizhao: Done /

PR #9, opened 2020-07-03
Get comments by video id in backend.

kayefung:
Given a video ID, ```getCommentsFromId()``` uses YouTube Data API to return a list of 25 comments from the video (ordered by relevance i.e. the same order as "Top comments" appear in the comment section of a video).

Also included tests for to check that an empty list is returned for videos where comments cannot be found (ex. video doesn't have comments, video is private and can't be accessed).


src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+      if (response.getItems().isEmpty()) {
+        throw new Exception(String.format("Video %s has no comments.", videoId));
+      }
+
+      List<String> commentsList = new ArrayList<>();
+      for (CommentThread commentThread : response.getItems()){
+        String commentContent = commentThread.getSnippet()
+            .getTopLevelComment()
+            .getSnippet()
+            .getTextDisplay();
```
nif33, line 58: NIT: indentation # of spaces is inconsistent here

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Unlisted videos should be accessible to the API if the video id is inputted. List of comments
+   * should be returned for unlisted videos just like they would for a public video. 
+   */
+  @Test
+  public void unlistedVideoReturnsCommentsList() throws GeneralSecurityException, IOException {
+    String videoId = "9BUFUUhfXSg"; // Unlisted test video that has 2 comments. 
+    List<String> actual = commentService.getCommentsFromId(videoId);
+
+    Assert.assertFalse(actual.isEmpty());
```
nif33, line 38: Maybe for this one you can hardcode the comments to check that each comment matches since its an unlisted video you own

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+    List<String> actual = commentService.getCommentsFromId(videoId);
+
+    Assert.assertEquals(null, actual);
+  }
+  
+  /**
+   * For videos that have disabled commenting, a GoogleJsonResponseException should be caught and
+   * null should be returned. 
+   */
+  @Test
```
nif33, line 57: great test! I did not think of this case

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+  public void noCommentsReturnsNull() throws GeneralSecurityException, IOException {
+    String videoId = "ERL0EDZT1kE";
+    List<String> actual = commentService.getCommentsFromId(videoId);
+
+    Assert.assertEquals(null, actual);
+  }
+
+  /**
+   * Invalid IDs that don't refer to a video should return null. 
+   */
```
nif33, line 78: should we test this case in the backend or in the frontend before we even send the data to the backend? or both??

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.youtube.YouTube;
+import com.google.api.services.youtube.model.CommentThread;
+import com.google.api.services.youtube.model.CommentThreadListResponse;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.lang.Exception;
```
chenzizhao, line 13: nit: import should follow ascii order (yeah I didn't know this exists either) 

https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * A list of Strings should be returned for any public video that has comments. 
+   */
+  @Test
+  public void publicVideoReturnsCommentsList() throws GeneralSecurityException, IOException {
+    String videoId = "v3whnmU619I"; // Public video that has some comments.
+    List<String> actual = commentService.getCommentsFromId(videoId);
+
+    Assert.assertFalse(actual.isEmpty());
```
chenzizhao, line 26: I think the test could be more specific, because right now the commentService.getCommentsFromId() could return a hard coded string instead of calling the api, and it still passes this test (but not other tests of course).

chenzizhao: same point as Jennifer's (lol)

src/test/java/com/google/sps/CommentServiceTest.java
chenzizhao: I think there are two types of bad ids:

1. Those that are apparently illegal without calling the api, like "", or "123", or contains illegal characters like "abc//abc", we can certainly check these type of error in the front end. 
2. Those that appear legal but aren't, like replacing a few characters in a proper video id. We have to call the api in this case.

My suggestion is that we can have a `isValidId` checker class in the back end before making other api calls. It will certainly be slower than if we check it in the front end (in the case of type 1 illegal id), but I think it is a good idea to  check the validity in one place.

This would benefit the `Caption` class too :)

src/test/java/com/google/sps/CommentServiceTest.java
chenzizhao: +1 I didn't even know it is possible to disable commenting 😅

src/main/java/com/google/sps/servlets/CommentService.java
kayefung: Yeah, looks a bit weird but it's part of the [style guide](https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent).

src/main/java/com/google/sps/servlets/CommentService.java
kayefung: Didn't know this, thanks!

src/test/java/com/google/sps/CommentServiceTest.java
kayefung: I agree with checking valid IDs in the front end, one thing I'll mention is there doesn't seem to be any pattern to the IDs so I don't know how we can define something like "123" as "apparently illegal".

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,90 @@
...
+   * Unlisted videos should be accessible to the API if the video id is inputted. List of comments
+   * should be returned for unlisted videos just like they would for a public video. 
+   */
+  @Test
+  public void unlistedVideoReturnsCommentsList() throws GeneralSecurityException, IOException {
+    String videoId = "9BUFUUhfXSg"; // Unlisted test video that has 2 comments. 
+    List<String> expected = Arrays.asList("test comment 2", "test comment 1");
+    List<String> actual = commentService.getCommentsFromId(videoId);
+
+    Assert.assertEquals(expected, actual);
```
chenzizhao, line 41: nit: I am not sure how "sort by relevance" is defined, but if "test comment 1" on your test video gets a lot of thumbs up, is the expected list in a different order and break your test?

kayefung: This is a really good point, and I think my logic for this test was flawed now that I think more about it. Originally I thought, since I unlisted this video, no one other than me could access it and be able to affect the comment section. But I failed to acknowledge that the video ID is publicly available on github so anyone could find it and break this test :/

A current solution may be using containsAll() and adjusting the video settings so only approved comments are posted. I feel like this isn't a very sufficient test though.

chenzizhao: It makes sense to me that by unlisting the video, people won't find the video as easily. And also I think this is less a coding/logic issue but more of an infrastructure one. I can't think of a good workaround either 🤷‍♀️

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Calls YouTube Data API to return a list of the top 25 comments for a video, identified by its
+   * video id. 
+   */
+  public List<String> getCommentsFromId(String videoId)
+      throws GeneralSecurityException, IOException, GoogleJsonResponseException {
+    YouTube youtubeService = getService();
+    YouTube.CommentThreads.List request = youtubeService.commentThreads()
+        .list(Collections.singletonList("snippet"));
+    try {
```
ajpaon, line 40: Can you make the try block smaller so it's clear which statements can throw an exception?

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+   */
+  public List<String> getCommentsFromId(String videoId)
+      throws GeneralSecurityException, IOException, GoogleJsonResponseException {
+    YouTube youtubeService = getService();
+    YouTube.CommentThreads.List request = youtubeService.commentThreads()
+        .list(Collections.singletonList("snippet"));
+    try { 
+      // Call YouTube API and get the 25 comments that appear first in comment section of the video
+      CommentThreadListResponse response = request.setKey(DEVELOPER_KEY)
+          .setMaxResults(25L)
```
ajpaon, line 43: Consider pulling this number out into a named constant

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+          .setOrder("relevance")
+          .setTextFormat("plainText")
+          .setVideoId(videoId)
+          .execute();
+      
+      if (response.getItems().isEmpty()) {
+        throw new Exception(String.format("Video %s has no comments.", videoId));
```
ajpaon, line 50: Why did you decide to throw an exception instead of returning an empty list?

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+      }
+  
+      return commentsList;
+
+    } catch (GoogleJsonResponseException e) {
+      System.out.println(e.getDetails().getMessage());
+      // If comments cannot be retrieved, return null to indicate that sentiment score should not
+      // be calculated. An arbitrary score of 0 will lower the average sentiment score between
+      // captions and comments scores. 
+      return null;
```
ajpaon, line 69: If we're going to return null from this function, be sure to mention it in the javadoc.

But IMO it's better to avoid returning null if possible. Because now all callers need to remember to check if the return value is null before using it (and if they don't, we'll end up crashing with NullPointerException eventually)

src/test/java/com/google/sps/CommentServiceTest.java
juanalonsocruz: If someone else can do the job for you and return to you a nice error then you shouldn't do it yourself :). Haven't read through your implementation yet but if you call the API with a bad video id and it tells you it is a bad id, just catch it and tell the client that they need to input a valid id.

src/test/java/com/google/sps/CommentServiceTest.java
juanalonsocruz: This is a good example of why unit test should be self contained and not actually call real services. You should be mocking the response to something hard coded so the test doesn't become flaky.

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+        .build();
+  }
+
+  /**
+   * Calls YouTube Data API to return a list of the top 25 comments for a video, identified by its
+   * video id. 
+   */
+  public List<String> getCommentsFromId(String videoId)
+      throws GeneralSecurityException, IOException, GoogleJsonResponseException {
+    YouTube youtubeService = getService();
```
juanalonsocruz, line 37: Nit: This method builds a service client every time it is called. Is it possible to only do this once? Not even sure if its useful to do so.

src/main/java/com/google/sps/servlets/CommentService.java
juanalonsocruz: If someone throws an exception because of a bad request, you should let the client know what went wrong by creating a HTTP error response.

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+      System.out.println(e.getDetails().getMessage());
+      // If comments cannot be retrieved, return null to indicate that sentiment score should not
+      // be calculated. An arbitrary score of 0 will lower the average sentiment score between
+      // captions and comments scores. 
+      return null;
+    }
+    catch (Exception e) {
+      System.out.println(e.getMessage());
+      return null;
+    }
```
juanalonsocruz, line 74: Generic Exception catches aren't recommended. Do you have a list of all the potential exceptions that this block can throw?  Do some of them need to be handled differently? Are some of them avoidable?

You can also specify catch blocks to catch any exceptions from a given list.

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,77 @@
...
+
+    } catch (GoogleJsonResponseException e) {
+      System.out.println(e.getDetails().getMessage());
+      // If comments cannot be retrieved, return null to indicate that sentiment score should not
+      // be calculated. An arbitrary score of 0 will lower the average sentiment score between
+      // captions and comments scores. 
+      return null;
+    }
+    catch (Exception e) {
+      System.out.println(e.getMessage());
```
juanalonsocruz, line 72: System.err.println might be preferred. IDEs show the text as red as well so nicer to visually catch.

src/main/java/com/google/sps/servlets/CommentService.java
juanalonsocruz: Just noticed this is not a HTTP servlet. In that case an illegalArgumentException should be thrown instead.

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,90 @@
...
+public final class CommentServiceTest {
+
+  CommentService commentService = new CommentService();
+
+  /**
+   * A list of Strings should be returned for any public video that has comments. 
+   */
+  @Test
+  public void publicVideoReturnsCommentsList() throws GeneralSecurityException, IOException {
+    String videoId = "v3whnmU619I"; // Public video that has some comments.
```
juanalonsocruz, line 25: General comment for all unit tests. You shouldn't be calling other services in unit test. Unit test should be self contained. You should mock the service calls to return hard coded responses. In practice, unit tests are run many times automatically making sure things are still working. You wouldn't want to waste your quota on unit tests.

src/main/java/com/google/sps/servlets/CommentService.java
kayefung: Yeah I think throwing an exception was pretty pointless now that I think about it... I guess I could write tests to check that the correct exception was thrown in different cases?

src/main/java/com/google/sps/servlets/CommentService.java
kayefung: I think GoogleJsonResponseException is thrown for all the errors that can happen, so we'll return an empty list in those cases. I decided to take out the generic exception

For cases where the video id isn't valid or the video is private/can't be accessed, we'll be checking for those cases in the front end, so do I still handle that case in the backend as well?

src/main/java/com/google/sps/servlets/CommentService.java
kayefung: ack, I tried to make a CommentService constructor that would call getService, but then whenever you create a CommentService object you'd have to throw or catch GeneralSecurityException and IOException. Doable but I'm not sure if it's worth it?

kayefung:
Added ```src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker``` so I could mock final classes (seems like everything I wanted to mock that was part of the YT API was final)

src/main/java/com/google/sps/servlets/CommentService.java
juanalonsocruz: Hmm.. What happens if instead of inside a constructor, you fill the object property when you declare it. I'm guessing you would have the same issue but interested if it didn't

src/main/java/com/google/sps/servlets/CommentService.java
juanalonsocruz: Can't assume any checks work in the front end. Anyone can send a POST GET request directly here without going via the front end code. I recommend you do a validation check in this servlet as well. You shouldn't make it yourself though. Either call another YouTube API service that validates it or if you the call you already make in this PR tells you that it is an invalid id, you should catch it and throw an error upwards saying it was an invalid argument.

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,109 @@
...
+    when(requestMock.setKey(anyString())
+        .setMaxResults(anyLong())
+        .setOrder("relevance")
+        .setTextFormat("plainText")
+        .setVideoId(anyString())
+        .execute())
+        .thenReturn(responseMock);
+    when(responseMock.getItems()).thenReturn(threadList);
+    when(commentThreadMock.getSnippet().getTopLevelComment().getSnippet().getTextDisplay())
+        .thenReturn("a", "b"); // Returns "a" first time, "b" second time.
```
juanalonsocruz, line 62: Hmm I'm not very familiar with Java mocking but I feel like this might be too many mocks. You want to minimize mocks only to when you call something outside this service. In this PR, that would be building the youtube api service and calling the list comment threads method. You are doing these two with the `when` call in the setup method and the first one here. The remaining `when` calls here are used to mock the response you already mocked. What you want to test is how your code handles the response from the API. Mocking anything after the response, means you might be faking too much. 

Is there a way to create an actual an actual CommentThread List object and then force to be returned in the first `when` in this state such that when it calls getItems, getSnippet, getTopLevelComment, getSnippet, and getTextDisplay it all works naturally?

Creating this object to be returned might involve a lot of setting up, so I recommend creating a helper method where the arguments are the relevant data needed for the test. In this case, the relevant data would be `"a"` and `"b"`.

Copied this internal doc for the pod. Mock Object Best Practices https://drive.google.com/file/d/1W9iZyu5kYOTUjO6_WHTNqUy4ie6FmHyc/view?usp=sharing

If you want to work on this together, feel free to set up some time on my calendar and you can screen share what you are doing. Meeting like this reduces the friction between PR reviews. I recommend you first attempt it yourself so there isn't a lot of waiting time during our meeting.

src/main/java/com/google/sps/servlets/CommentService.java
kayefung: I think the same issue would apply, but I ended up using a constructor anyway haha

src/test/java/com/google/sps/CommentServiceTest.java
kayefung: Thanks for the offer! I think I figured out what you meant, your description was super helpful :)

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,79 @@
...
+          .setMaxResults(MAX_RESULTS)
+          .setOrder("relevance")
+          .setTextFormat("plainText")
+          .setVideoId(videoId)
+          .execute();
+    } catch (GoogleJsonResponseException e) {
+      System.err.println(e.getDetails().getMessage());
+      
+      for (GoogleJsonError.ErrorInfo err : e.getDetails().getErrors()) {
+        if (!err.getReason().equals("commentsDisabled")) { // Ex. invalid video id or private video
```
juanalonsocruz, line 60: Great! You can differentiate between commentsDisabled and not. Could you also differentiate between invalid/private video and any other issue? Asking because the former is the callers fault (need to throw IllegalArgumentException) and the latter is not the callers fault (need to throw some kind of internal exception. There isn't a built in exception for this one so I would just rethrow the same exception.)

juanalonsocruz: Remembered again this is not a servlet. If it was these error differentiation is useful so you could return the correct HTTP error response. But since its not a servlet, I guess you could just not catch the errors. I'm hesitant to not catch the errors because the caller of this method will be coming from a servlet which would still need to do this differentiation. 

Why is this not a servlet? Who will be calling this method?

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -0,0 +1,79 @@
...
+          throw new IllegalArgumentException();
+        } else { // Video with disabled comments
```
juanalonsocruz, line 62: No need for an else if you are returning inside the else. 
Not Preferred
```
if (...) {
  return/throw A
} else {
  return/throw B
}
```

Preferred
```
if (...) {
  return/thow A
}

return/throw B
```

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,179 @@
...
+  @Mock CommentThreadListResponse responseMock;
+  @Mock(answer = Answers.RETURNS_DEEP_STUBS) YouTube youtubeServiceMock;
+  @Mock(answer = Answers.RETURNS_DEEP_STUBS) YouTube.CommentThreads.List requestMock;
+  @Mock(answer = Answers.RETURNS_DEEP_STUBS) GoogleJsonResponseException exceptionMock;
+
+  @InjectMocks CommentService commentService;
+
+  @Before
+  public void setUp() throws IOException {
+    when(youtubeServiceMock.commentThreads().list(any())).thenReturn(requestMock);
```
juanalonsocruz, line 46: Nit: Remove extra line.

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,179 @@
...
+    List<String> actual = commentService.getCommentsFromId(VIDEO_ID);
+
+    Assert.assertEquals(Arrays.asList(), actual);
+  }
+
+  /**
+   * Creates a list of CommentThread objects that stores comment texts specified by the parameter.
+   * The returned list is set as the items of a CommenThreadListResponse. 
+   */
+  private List<CommentThread> createCommentThreadList(List<String> strList) {
```
juanalonsocruz, line 143: NICE!

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -0,0 +1,179 @@
...
+  @Test
+  public void videoReturnsCommentsList() throws GeneralSecurityException, IOException {
+    CommentThreadListResponse response = new CommentThreadListResponse();
+    response.setItems(createCommentThreadList(new ArrayList<>(Arrays.asList("a", "b"))));
+
+    when(requestMock.setKey(anyString())
+        .setMaxResults(anyLong())
+        .setOrder("relevance")
+        .setTextFormat("plainText")
+        .setVideoId(anyString())
```
juanalonsocruz, line 61: Switch anyString() with VIDEO_ID

src/main/java/com/google/sps/servlets/CommentService.java
kayefung: I tested it out with a few videos and noticed that getReason() is the same for invalid ids and private videos (both just "videoNotFound") :( I'll separate it out so that it throws IllegalArgumentException for videoNotFound, returns an empty list for commentsDisabled, and then rethrows the same exception for anything else for now?

SentimentServlet is supposed to get the user input for the video id, call this method, get comments (and captions) and then perform sentiment analysis on the data.

src/test/java/com/google/sps/CommentServiceTest.java
kayefung: I think this is just github's line wrapping lol

src/main/java/com/google/sps/servlets/CommentService.java
```diff
@@ -57,11 +76,12 @@ private YouTube getService() throws GeneralSecurityException, IOException {
...
       for (GoogleJsonError.ErrorInfo err : e.getDetails().getErrors()) {
-        if (!err.getReason().equals("commentsDisabled")) { // Ex. invalid video id or private video
-          throw new IllegalArgumentException();
-        } else { // Video with disabled comments
+        if (err.getReason().equals("videoNotFound")) { // Invalid video id or private video
+          throw new IllegalArgumentException("Video is private or does not exist.");
+        } else if (err.getReason().equals("commentsDisabled")) { // Video with disabled comments
           return new ArrayList<>();
         }
+        throw e;
```
juanalonsocruz, line 84: Move this after the for loop. Maybe other errors in the error list can have video not found or comments disabled.

src/test/java/com/google/sps/CommentServiceTest.java
```diff
@@ -115,6 +115,28 @@ public void disabledCommentsReturnsEmptyList() throws GeneralSecurityException,
...
+            "Other error detail (not commentsDisabled or videoNotFound)");
+    
+    when(requestMock.setKey(anyString())
+        .setMaxResults(anyLong())
+        .setOrder("relevance")
+        .setTextFormat("plainText")
+        .setVideoId(VIDEO_ID)
+        .execute())
+        .thenThrow(exceptionMock);
+    when(exceptionMock.getDetails()).thenReturn(error);
```
juanalonsocruz, line 135: Is it possible to set details to be the error instead? This one of those cases where you mocking more than what you need to.

kayefung: ack but leaving as is because creating a new GoogleJsonResponseException requires making a lot of other new objects

PR #10, opened 2020-07-03
add tests to Perspective API

nif33:
refactored the API call: created a class for comments so that I can test all methods related to preparing a new comment for the API call.
Using QUnit tests: added tests.html file which displays the qunit tests


src/main/webapp/scripts/script.js
```diff
@@ -0,0 +1,6 @@
+const API_KEY = '';
+
+function updateDom(text, containerId) {
+  const containerDOM = document.getElementById(containerId);
+  containerDOM.innerText = text;
+}
```
nif33, line 6: I added this to script.js because several pages might want this helper function and to access the API key

src/main/webapp/scripts/tests/comment-class-test.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+QUnit.test('Comment Class - getError (input size exceeded)', function(assert) {
+  const comment = new Comment('A'.repeat(3001));
+  const result = comment.getError();
+  const expected = 'Comment is too long. We are currently only able to analyze comments of up to approx. 3000 characters.';
+  assert.equal(result, expected, "error message for comment length exceeded");
+});
+
+QUnit.test('Comment Class - makePerspectiveRequestString (normal use)', function(assert) {
```
nif33, line 22: makePerspectiveRequestString kinda requires that the programmer check comment.isValid first... I could give it a default comment if isValid is false but this shouldn't even be called if it's false... idk if you have a better idea, or more edge cases to test, please let me know!

src/main/webapp/scripts/perspective.js
```diff
@@ -1,21 +1,30 @@
...
   const request = new Request(
     `https://commentanalyzer.googleapis.com/v1alpha1/comments:analyze?key=${API_KEY}`,
-    {method: 'POST', body: json}
+    {
+      method: 'POST',
+      body: comment.perspectiveString;
+    }
   );
-  fetch(request)
+  fetch(comment.perspectiveRequest)
```
chenzizhao, line 20: Should be request?

src/main/webapp/scripts/tests/comment-class-test.js
```diff
@@ -0,0 +1,27 @@
+QUnit.test('Comment Class - getError (normal use)', function(assert) {
```
chenzizhao, line 1: nit: If QUnit supports that, is it possible to move the test files to src/test/webapp/script/...? That way test code and main code can be separated

src/main/webapp/scripts/script.js
chenzizhao: Yap looks helpful!

src/main/webapp/scripts/perspective.js
```diff
@@ -1,21 +1,30 @@
...
     .then(response => response.json())
     .then((responseJson) => {
       const score = responseJson.attributeScores.TOXICITY.summaryScore.value;
-      const containerDOM = document.getElementById('toxicity-container');
-      containerDOM.innerText = (score*100).toFixed(2) + '% likely to be toxic.';
+      updateToxicityDom((score*100).toFixed(2), 'toxicity-container');
+    },
+    function(err) {
```
chenzizhao, line 26: nit: I would probably use catch(error)=>{...} here, because it explicitly shows this is a try-catch block, and also consistent with other arrow functions in getToxicity()

src/main/webapp/scripts/tests/comment-class-test.js
nif33: Yes, but it would result in some tags like "../../../test/webapp/script/" because the tests require both js files in the script directory and test directory. Might be easy to get confused with all the ../, what do you think?

The other option is to use an absolute file path "/src/test/webapp/script/.." and "/src/webapp/script" but I read somewhere that relative file paths are preferred.

chenzizhao: I always thought absolute paths is preferred lol. It's okay, not a big deal. No need to change here. 

Does Qunit run automatically after `mvn appengine:run` is called?

nif33: Hmm googled it and found some mixed answers, the style guide tends to use relative paths though, the main reason I found for relative paths is that its easier to change file names but I don't think we'll be needing to do that. as long as we stick with one way it should be fine

Qunit runs when you load the website at /test.html, I guess from now on we'll have to check that page before every push. Me and Kaye did not find a JS testing framework that was convenient to use with mvn test :/

chenzizhao: these are ways to connect them all (qunit or other testing frame works, node and mvn) but it seems like a pain from what I have found (this is probably what cicd folks do?). I agree right now we can just manually check the test page.

nif33: You're right, I just searched and found a couple options but some issues I encountered were
-could not find documentation
-licenses
-broken links
Probably better to just use the test page and look into this again later if needed. Or if you find one that might work, link me and I'll look into it!

src/main/webapp/scripts/perspective.js
```diff
@@ -1,21 +1,30 @@
...
   fetch(request)
     .then(response => response.json())
     .then((responseJson) => {
       const score = responseJson.attributeScores.TOXICITY.summaryScore.value;
-      const containerDOM = document.getElementById('toxicity-container');
-      containerDOM.innerText = (score*100).toFixed(2) + '% likely to be toxic.';
+      updateDom((score*100).toFixed(2), 'toxicity-container');
+    })
+    .catch((err) => {
+      alert("The analysis feature seems to be unavailable at this moment.");
```
ajpaon, line 27: does it make sense to put this message into the DOM instead of an alert? alerts can be a bit annoying.

src/main/webapp/scripts/perspective.js
```diff
@@ -1,21 +1,30 @@
...
+      body: comment.perspectiveString
+    }
   );
   fetch(request)
     .then(response => response.json())
     .then((responseJson) => {
       const score = responseJson.attributeScores.TOXICITY.summaryScore.value;
-      const containerDOM = document.getElementById('toxicity-container');
-      containerDOM.innerText = (score*100).toFixed(2) + '% likely to be toxic.';
+      updateDom((score*100).toFixed(2), 'toxicity-container');
```
ajpaon, line 24: this formula isn't totally obvious. consider separating out a helper function that takes a number and returns a percentage

src/main/webapp/scripts/perspective.js
```diff
@@ -1,21 +1,30 @@
...
-  const commentText = document.getElementById("comment-text").value;
-  const commentObject = {
-    'comment': {'text': commentText},
-    'requestedAttributes': {'TOXICITY': {}},
-    'languages': ['en']
-  };
-  const json = JSON.stringify(commentObject);
+  const comment = new Comment(document.getElementById("comment-text").value);
+  if(!comment.isValid) {
+    updateDom(comment.getError(), 'toxicity-container');
```
ajpaon, line 10: It's kind of a shame that we need to call getError twice whenever a Comment is invalid (once in the constructor and once here). Also, every client of Comment is responsible for checking that it is valid before using it. Any ideas about how to avoid these issues?

src/main/webapp/scripts/classes/comment-class.js
```diff
@@ -0,0 +1,32 @@
...
+  #MAX_COMMENTS  = 3000;
+  #NO_COMMENT_ERROR = 'No comment inputted';
+  #COMMENT_LENGTH_EXCEEDED_ERROR = 'Comment is too long. We are currently only able to analyze comments of up to approx. 3000 characters.';
+  constructor(text) {
+    this.text = text;
+    this.isValid = !this.getError();
+    this.perspectiveString = this.makePerspectiveRequestString();
+  }
+  getError() {
+    if(this.text == 0) {
```
juanalonsocruz, line 16: `text` seems like a string type and it is being compared to `0`. Is this intended?

src/main/webapp/scripts/perspective.js
nif33: added this function and a test for it, I kinda messed up my git tree during that though so sorry about the confusing git history

src/main/webapp/scripts/classes/comment-class.js
```diff
@@ -0,0 +1,32 @@
...
+      return this.#NO_COMMENT_ERROR;
+    }
+    if(this.text.length > this.#MAX_COMMENTS) {
+      return this.#COMMENT_LENGTH_EXCEEDED_ERROR;
```
juanalonsocruz, line 20: Returning strings for errors seems odd. How about throw JS errors? Or not even letting the constructor finish creating the comment object and throwing an error telling the caller that they called it incorrectly?

src/main/webapp/scripts/tests/comment-class-test.js
juanalonsocruz: What do you mean by it requiring it? If you feed it an invalid string what happens? does it give a nice response saying what you did wrong? If so, thats not bad (I think).

src/main/webapp/scripts/classes/comment-class.js
nif33: I did this to check if the string is empty because non-empty strings are true, but I realized that there are [a lot of ways ](https://stackoverflow.com/questions/154059/how-can-i-check-for-an-empty-undefined-null-string-in-javascript)to do this. I will change it to !this.text because I think that makes more sense to look at than the integer comparison?

src/main/webapp/scripts/classes/comment-class.js
nif33: I changed it to throw errors and the code looks a lot better, thanks for the suggestion! I did not think of making my own error objects

src/main/webapp/scripts/perspective.js
nif33: avoided by having the constructor throw errors instead of having a function that returns strings

src/main/webapp/scripts/classes/comment-class.js
juanalonsocruz: A common way is to do `!this.text`.

nif33: I did change it, is the new version not showing up? (I messed up my git history at some point so maybe you'll have to toggle the "Changes from ___ commits" button in the Files changed tab?)

src/main/webapp/scripts/perspective.js
```diff
@@ -4,14 +4,19 @@
...
+   return (toxicity*100).toFixed(2).toString() + '%';
  }
 
 function getToxicity() {
-  const comment = new Comment(document.getElementById("comment-text").value);
-  if(!comment.isValid) {
-    updateDom(comment.getError(), 'toxicity-container');
+  let comment;
+  try {
+    comment = new Comment(document.getElementById('comment-text').value);
```
juanalonsocruz, line 17: I actually don't know if this is good practice or not. To me it seems a little bit odd that you are sharing data between JS files through the DOM (at least that's what it looks you will be doing once you have the comments API ready). 

I think we want to adapt a design model for organizing your code (ex is Model View Control model). Basically you want one or more script files that gather data via APIs and another main script file that imports the API script files. The main script file's role is to get data from the other script files and manipulate the DOM. With this you would only have one script tag in the HTML for the main script file.

For your case, I think a Comment object would contain all the data the UI needs to display (for now it seems like its just the comment text and the toxicity). This means you would need to call the comments API to get the comments, then call the Perspective API to get the toxicity, then build the Comment object, and then pass the object to method that knows how to use the properties in that object to display it in the UI.

Does this makes? I feel like I'm even confusing myself.

https://stackoverflow.com/questions/950087/how-do-i-include-a-javascript-file-in-another-javascript-file
https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller

nif33: These functions are for the Perspective API so we are not sharing data using the DOM (getting youtube comments is completely in the backend). The 'comment-text' ID is for user-inputted new comments which are the only type of comment that will be handled in the front end code. I could change the class name to something like userComment, but the two uses of comments are completely separated by frontend/backend so is it still necessary?

I refactored such that:
-Model: perspective-api-service.js - accepts requests for api call, responds with api data (toxicity)
-Model: comment-class - accepts requests to make comment object, responds with comment object that is ready for API call
-View: script.js - functions that manipulate the DOM
-Controller: perspective.js - handles loading relevant scripts, contains function that the user input button calls, sends data to user interface functions

From your description, I realize that scripts should be loaded in the View, but I don't think that makes sense for our project since we want to use the same UI functions in many html pages, but not the same js scripts. Therefore, I put script loading in the Controller instead (for now).
Also, I realize the View should be doing more work (handling an object and deciding how to display it) but since there is so little data to display right now (1 line of text) I think it would make sense to implement that later on when there is more data.

I hope this has made it easier to read, it was very fun to implement the MVC model! I imagine it'll be very helpful when Zoe adds the submit comment to Youtube feature

src/main/webapp/scripts/perspective.js
```diff
@@ -1,21 +1,26 @@
...
+  try {
+    comment = new Comment(getDomValue('comment-text'));
+  } catch(error) {
+    updateDom(error.message, 'toxicity-container');
+  }
+  perspectiveApiCall(comment, function(error, result) {
+    if(error != null) {
+      updateDom(error.message, 'toxicity-container');
+      return;
+    }
```
nif33, line 19: Handling errors in callback functions is kinda weird to me, but I wanted to stick with the pattern of this file handing calling the DOM functions. I found [this](https://medium.com/@1sherlynn/how-to-handle-errors-in-asynchronous-javascript-code-when-working-with-callbacks-dcd32bca4b6b) article that told me to use this pattern, which is a familiar pattern from NodeJS projects I have worked on. Let me know if there are better preferred patterns though

src/main/webapp/scripts/perspective-api-service.js
```diff
@@ -0,0 +1,24 @@
+/*
+ *  Calls the Perspective API
+ *  callback function is called after API call completed
+ */
+
+function perspectiveApiCall(comment, callback) {
```
juanalonsocruz, line 6: Not very familiar with practices on using callback. Where did you find examples like this one? Why not make it just a function that returns the toxicity and then let the caller to whatever they want with it?

https://stackoverflow.com/questions/950087/how-do-i-include-a-javascript-file-in-another-javascript-file

It seems that you need to export the method in this file by doing `export function getToxicity(comment) {...}` and then in another js file you can do `import { getToxicity } from '<path to perspective-api-service.js>'` and then now you can call getToxicity("this is a comment"). With this you wouldn't lead to load all js files in the DOM. Only the main js.

Please let me know if you found good practices that prefer otherwise :)

juanalonsocruz: Also, if you want to discuss this at a faster rate. Maybe we could have GVC talk. This reduces a lot of friction between PR comments. If we were working together physically this is something I would schedule some time to discuss. Throughout the summer, I would prefer we get in quick gvc chats to discuss these topics quicker.

nif33: Zoe showed me how she used the `import { function } from '<path>'` method, but I don't think its suitable for this file because it makes the perspective.js file into a module, meaning that it's functions cannot be called from elsewhere in the code. Since getToxicity() may be called from many places such as the chrome extension, I think it's best to leave it as is.

src/main/webapp/scripts/classes/comment-class.js
```diff
@@ -0,0 +1,29 @@
...
+    this.text = text;
+    if(!this.text) {
+      console.error("User did not input any comment");
+      throw NO_COMMENT_ERROR;
+    }
+    if(this.text.length > this.#MAX_COMMENTS) {
+      console.error("User exceeded maximum comment length");
+      throw COMMENT_LENGTH_EXCEEDED_ERROR;
+    }
+    this.perspectiveString = this.makePerspectiveRequestString();
```
juanalonsocruz, line 19: I find `perspectiveString` to not be a property a comment. I would expect more to see a toxicity property. So maybe something like this.toxicity = <perspective_api_js_file>.getToxicity(this.text). This getToxicity() method knows about the perspective request body because its in the perspective api file. comment class file shouldn't know details about the perspective api. 

After this change, now you have a nice comment object with properties that you care about and in the `view` file from `MVC`  you know to use this comment object (displaying that this comment text has this comment toxicity).

src/main/webapp/scripts/tests/comment-class-test.js
```diff
@@ -0,0 +1,34 @@
+QUnit.test('Comment Class - Constructor (normal use)', function(assert) {
```
juanalonsocruz, line 1: In this file you should be testing that when you create a comment, it handles the error cases you made and you should also be testing that it gets a toxicity number after mocking the API.

src/main/webapp/scripts/tests/perspective-test.js
```diff
@@ -1,4 +1,41 @@
...
+  });
+});
+
+QUnit.test('getToxicityFromPerspectiveApi', function(assert) {
+  const done = assert.async();
+  const comment = 'normal comment';
+  sinon.stub(window, 'fetch');
+  window.fetch.returns(Promise.resolve(mockPerspectiveResponse));
+  getToxicityFromPerspectiveApi().then((response) => {
+    assert.equal(response, 0, 'score is 0');
```
juanalonsocruz, line 32: Nit: Good practice to show in the test that you create mocked response with specific values you are expecting. In this case, it should be something like

```
window.fetch.returns(Promise.resolve(createMockPerspectiveResponse(TOXICITY)));
getToxicityFromPerspectiveApi().then((response) => {
    assert.equal(response, TOXICITY);
```

PR #13, opened 2020-07-13
Post a comment in the front end

chenzizhao:
This PR included a function in comments.js that makes a POST request to the YouTube server. It returns nothing if successful and an error if failed. 

* Use QUnit to test (tests.html)
* Use sinon.js to stub responses

### DIFF
--- a/src/main/webapp/scripts/post-comment.js
+++ b/src/main/webapp/scripts/post-comment.js
@@ -1,0 +1,27 @@
+function postVideoComment(videoId, commentText) {
+    return new Promise((resolve, reject) => {
+        gapi.client.request(buildRequestBody(videoId, commentText))
+            .then(res => res.json)
+            .then(data => resolve(data))
+            .catch(err => reject(new Error(err.result.error.message)));
+    });
+}
+
+function buildRequestBody(videoId, commentContent) {
+    return {
+        'path': '/youtube/v3/commentThreads?part=snippet',
+        'method': 'POST',
+        'body': {
+            'snippet': {
+                'videoId': videoId,
+                'topLevelComment': {
+                    'snippet': {
+                        'textOriginal': commentContent
+                    },
+                },
+            },
+        },
+    };
+}
+
+export { postVideoComment };
--- a/src/main/webapp/scripts/tests/comment-post-test.js
+++ b/src/main/webapp/scripts/tests/comment-post-test.js
@@ -1,0 +1,95 @@
+import { postVideoComment } from '/scripts/post-comment.js';
+
+const VIDEO_ID = 'sample video id';
+const COMMENT_TEXT = 'sample comment';
+
+QUnit.module('Post a comment via YouTube API', {
+    beforeEach: () => {
+        gapi = new CustomGapi();
+        sinon.stub(gapi.client, 'request');
+    },
+    afterEach: () => { gapi.client.request.restore(); },
+});
+
+QUnit.test('Call API with correct request body', function (assert) {
+    const done = assert.async();
+
+    gapi.client.request.returns(Promise.resolve(new CustomResponse()));
+
+    postVideoComment(VIDEO_ID, COMMENT_TEXT)
+        .then(() => {
+            assert.deepEqual(gapi.client.request.getCall(0).args[0],
+                createExpectedRequestBody(VIDEO_ID, COMMENT_TEXT));
+            done();
+        });
+});
+
+QUnit.test('Relay the error with correct message', function (assert) {
+    const done = assert.async();
+    const errorMsg = 'error message';
+
+    gapi.client.request.returns(Promise.reject(new CustomError(errorMsg)));
+
+    postVideoComment(VIDEO_ID, COMMENT_TEXT)
+        .catch(error => {
+            assert.equal(error.message, errorMsg);
+            done();
+        });
+});
+
+QUnit.test('Successful post returns empty response', function (assert) {
+    const done = assert.async();
+    assert.expect(0);
+
+    gapi.client.request.returns(Promise.resolve(new CustomResponse()));
+
+    postVideoComment(VIDEO_ID, COMMENT_TEXT)
+        .then(() => {
+            done();
+        });
+});
+
+
+// Helper functions and classes
+
+class CustomResponse extends Response {
+    constructor() {
+        super();
+        this.json = {};
+    }
+}
+
+class CustomError extends Error {
+    constructor(msg) {
+        super();
+        this.result = { 'error': { 'message': msg, } };
+    }
+}
+
+class CustomGapi {
+    constructor() {
+        this.client = new CustomClient();
+    }
+}
+
+class CustomClient {
+    constructor() { }
+    request() { }
+}
+
+function createExpectedRequestBody(videoId, commentText) {
+    return {
+        'path': '/youtube/v3/commentThreads?part=snippet',
+        'method': 'POST',
+        'body': {
+            'snippet': {
+                'videoId': videoId,
+                'topLevelComment': {
+                    'snippet': {
+                        'textOriginal': commentText,
+                    },
+                },
+            },
+        },
+    };
+}
--- a/src/main/webapp/tests.html
+++ b/src/main/webapp/tests.html
@@ -16,5 +16,7 @@
   <script src="scripts/perspective-api-service.js"></script>
   <script src="scripts/classes/comment-class.js"></script>
   <script src="scripts/tests/perspective-test.js"></script>
+  <script src="https://apis.google.com/js/api.js"></script>
+  <script defer type="module" src="scripts/tests/comment-post-test.js"></script>
 </body>
 </html>


src/main/webapp/index.html
```diff
@@ -33,6 +33,18 @@
...
+  <div>
+    <p>Post a comment to </p>
+    <a href="https://www.youtube.com/watch?v=gjKpAAezWQA">this video</a>
+    <button id="authorize-button" style="display: none;">Authorize</button>
+    <button id="signout-button" style="display: none;">Sign Out</button>
+    <button id="comment-button" style="display: none;">Post a comment</button>
+    <div id="result-text"></div>
+  
+    <script defer type="module" src="scripts/controller.js"></script>
+    <script src="https://apis.google.com/js/api.js"></script>
```
chenzizhao, line 46: This is the part I mentioned in the sync about loading external resources. Here is the original snippet: https://github.com/google/google-api-javascript-client/blob/master/samples/authSample.html (line 117-119)

I tried the way they loaded authSample.js, but the console throws "can't find gapi" in controller.js. Maybe it is about the module type? 
```
    <script type="module" src="scripts/controller.js"></script>
    <script async defer src="https://apis.google.com/js/api.js"></script>
```

The way it is right now works, but delays a bit before all UI elements load. authSample.html has a slight delay too. Maybe not high priority but I wondered if there is any way to remove the delay?

nif33: I've seen the "can't find gapi" error before when loading the client, maybe we can have a GVC call to try some of the ways that worked for me, and maybe one will work for authentication as well?

src/main/webapp/scripts/controller.js
```diff
@@ -0,0 +1,47 @@
+import { makeApiCall } from '/scripts/comments.js';
+import { authorizeButton, signoutButton, commentButton, videoId, commentElement } from '/scripts/view.js';
+import { successfulApiCallback, failedApiCallback, updateSigninStatus } from '/scripts/view.js';
```
nif33, line 3: how did you get this part working? I'm unable to call functions in modules, can you teach me :grinning:

src/main/webapp/scripts/view.js
```diff
@@ -0,0 +1,36 @@
+export { authorizeButton, signoutButton, commentButton, videoId, commentElement, resultContainer };
+export { successfulApiCallback, failedApiCallback, updateSigninStatus }
+
+function updateDom(text, containerDOM) {
+    containerDOM.innerHTML = '';
```
nif33, line 5: are you using this line to empty the container? It may be unsafe to use innerHTML, instead you can use a loop that's like while there's children of the element, remove that child. I use this method in other code so it may be useful to have that as a general helper function in script.js

src/main/webapp/scripts/view.js
```diff
@@ -0,0 +1,36 @@
...
+
+function updateSigninStatus(isSignedIn) {
+    if (isSignedIn) {
+        authorizeButton.style.display = 'none';
+        signoutButton.style.display = 'block';
+        commentButton.style.display = 'block';
+    } else {
+        authorizeButton.style.display = 'block';
+        signoutButton.style.display = 'none';
+        commentButton.style.display = 'none';
```
nif33, line 25: I did something similar in my starter project with changing the style in JS files. It is not recommended, so maybe you could do something like change the button function instead? like 
```
const authButton = document.getElementById('auth-button');
if(isSignedIn) {
  authButton.onclick = handleSignoutClick;
  addCommentButton() //adds a comment button to the page
} else {
  authButton.onclick = handleAuthClick;
}
```
I'm not sure if this is the right way to avoid changing the style in JS, I'm sure there are many options that you can choose from

src/main/webapp/index.html
chenzizhao: yap talk to you after the sync?

src/main/webapp/scripts/controller.js
chenzizhao: sure!

src/main/webapp/scripts/view.js
chenzizhao: I don't see why it is unsafe since no user input is involved here with innerHtml, but I did find that while loop is faster at clearing child nodes. I will update that

src/main/webapp/index.html
chenzizhao: notes from the chat: 
- async = download the script from online sources
- defer = parse/run the script after the html are parsed and displayed

src/main/webapp/scripts/view.js
chenzizhao: Done

src/main/webapp/index.html
chenzizhao: moved to #14

src/main/webapp/scripts/comments.js
```diff
@@ -0,0 +1,25 @@
+export { makeApiCall, buildRequestBody };
+
+function makeApiCall(videoId, commentText, successfulApiCallback, failedApiCallback) {
```
juanalonsocruz, line 3: Needs more descriptive method name. `postVideoComment` or something like that.

src/main/webapp/scripts/comments.js
```diff
@@ -0,0 +1,25 @@
+export { makeApiCall, buildRequestBody };
+
```
juanalonsocruz, line 2: Should the file name be post-comment? Comparing to test file.

src/main/webapp/scripts/comments.js
```diff
@@ -0,0 +1,25 @@
+export { makeApiCall, buildRequestBody };
```
juanalonsocruz, line 1: Nit: I believe exports are usually found in the bottom of the file.

src/main/webapp/scripts/comments.js
juanalonsocruz, line 3: Question: Why use callbacks? Why not return the promise and let the caller do whatever they want to do with it?

```
    return gapi.client.request({
        'path': '/youtube/v3/commentThreads?part=snippet',
        'method': 'POST',
        'body': buildRequestBody(videoId, commentText),
    })
        .then(res => res.json)
```

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,69 @@
...
+const VIDEO_ID = 'sample video id';
+const COMMENT_TEXT = 'sample comment';
+
+QUnit.module('Post a comment', {
+    before: loadClientAsync,
+    beforeEach: () => { sinon.stub(gapi.client, 'request'); },
+    afterEach: () => { gapi.client.request.restore(); },
+});
+
+QUnit.test('BuildRequestBody', function (assert) {
```
juanalonsocruz, line 12: You should be testing behavior of the module and not helper methods. There is a ToTT for this. You should be testing that given a video id and comment text to makeApiCall, it make the API call with the correct request body, its mocked response is converted to the expected returned value of the method.

src/main/webapp/scripts/comments.js
chenzizhao: Done

src/main/webapp/scripts/comments.js
chenzizhao: Import at the beginning and export at the end I see I see. Done

src/main/webapp/scripts/comments.js
chenzizhao: Done.

src/main/webapp/scripts/comments.js
chenzizhao: To get the exact error message, the consumer has to go to `err.result.error.message`. I wanted to encapsulate this structure within `postVideoComment()` so that view.js don't have to worry about the structure of error message. That was one motivation I chose callback over promise.  Now I managed to use promise and return the exact error message :D

src/main/webapp/scripts/tests/comment-post-test.js
chenzizhao: yap done! (I miss how java uses private functions to enforce that practice)

src/main/webapp/scripts/post-comment.js
```diff
@@ -0,0 +1,27 @@
+function postVideoComment(videoId, commentText) {
+    return new Promise((resolve, reject) => {
+        gapi.client.request(buildRequestBody(videoId, commentText))
+            .then(res => res.json)
+            .then(data => resolve(data))
+            .catch(err => reject(err.result.error.message));
```
juanalonsocruz, line 6: Could you instead reject a custom error object with the error message? Strings aren't consider errors.

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,77 @@
...
+QUnit.test('Successful post', function (assert) {
+    var done = assert.async();
+
+    gapi.client.request.withArgs(REQUEST_BODY).returns(new Promise((resolve) => {
+        resolve(new CustomResponse());
+    }));
+
+    postVideoComment(VIDEO_ID, COMMENT_TEXT)
+        .then(data => {
+            assert.ok(true, `Don't care about the message`);
```
juanalonsocruz, line 50: I guess this line is not needed. All that matters is that the the promised was resolved thus just reaches done()

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,77 @@
...
+        .catch(error => {
+            assert.equal(error, errorMsg, 'Pass the error message to the callback function');
+            done();
+        });
+});
+
+QUnit.test('Successful post', function (assert) {
+    var done = assert.async();
+
+    gapi.client.request.withArgs(REQUEST_BODY).returns(new Promise((resolve) => {
```
juanalonsocruz, line 44: This might look repetitive and odd because of the file is small and too complex but you should create a helper test function that creates the expected request body that takes in video id and comment. You want to do this because you want to check "Given that I call postVideoComment with VIDEO_ID and COMMENT_TEXT, was gapi called with a request body that contains VIDEO_ID and COMMENT_TEXT in the correct fields?" You want this to be clear in the test itself without a reader going back up to check what REQUEST_BODY is.

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,77 @@
...
+    },
+};
+
+QUnit.module('Post a comment', {
+    before: loadClientAsync,
+    beforeEach: () => { sinon.stub(gapi.client, 'request'); },
+    afterEach: () => { gapi.client.request.restore(); },
+});
+
+QUnit.test('Throw an error', function (assert) {
```
juanalonsocruz, line 26: Test names should be more descriptive. Comment post failure returns error.

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,77 @@
...
+    }));
+
+    postVideoComment(VIDEO_ID, COMMENT_TEXT)
+        .catch(error => {
+            assert.equal(error, errorMsg, 'Pass the error message to the callback function');
+            done();
+        });
+});
+
+QUnit.test('Successful post', function (assert) {
```
juanalonsocruz, line 41: Successful comment post returns empty successful response.

src/main/webapp/scripts/post-comment.js
```diff
@@ -0,0 +1,27 @@
...
+    });
+}
+
+function buildRequestBody(videoId, commentContent) {
+    return {
+        'path': '/youtube/v3/commentThreads?part=snippet',
+        'method': 'POST',
+        'body': {
+            "snippet": {
```
FuzzyTree, line 15: nit: please use single or double quotes consistently here and throughout the code

src/main/webapp/scripts/tests/comment-post-test.js
chenzizhao: This is the error I get if I remove this line: 
`Expected at least one assertion, but none were run - call expect(0) to accept zero assertions.` I will call `expect(0)` instead

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,77 @@
...
+                "snippet": {
+                    "textOriginal": COMMENT_TEXT,
+                },
+            },
+        },
+    },
+};
+
+QUnit.module('Post a comment', {
+    before: loadClientAsync,
```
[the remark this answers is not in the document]
chenzizhao: Error ` Trying to stub property 'request' of undefined` if I comment this line out : /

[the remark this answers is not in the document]
juanalonsocruz: Are you loading the client in the source code? Wondering if this code to post a comment would work in production. I think it would be undefined error similar to what you have here. I believe Jennifer got this working already. You could talk to her.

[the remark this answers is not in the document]
chenzizhao: This code alone doesn't work because I loaded the client when the user authenticate themselves (#14 ). It only loads the client once instead of every time the user wants to post.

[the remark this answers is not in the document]
juanalonsocruz: Ah I see. The inner works of gapi.load are unknown to this file but it seems like it initializes gapi.client. I recommend you mock gapi.client so then you can mock gapi.client.request. Then you shouldn't have that undefined error without actually calling gapi.load :)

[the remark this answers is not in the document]
chenzizhao: Yah that sounds like a good plan. I know mocking is slightly different than stubbing in Sinon.js, and I haven't got a chance to look into the details. Can I make changes based on other comments first and come back to this?

src/main/webapp/scripts/post-comment.js
chenzizhao: lol fair point I will use reject(Error(...))

src/main/webapp/scripts/post-comment.js
chenzizhao: Fixed, nice catch :D

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,77 @@
...
+    beforeEach: () => { sinon.stub(gapi.client, 'request'); },
+    afterEach: () => { gapi.client.request.restore(); },
+});
+
+QUnit.test('Throw an error', function (assert) {
+    var done = assert.async();
```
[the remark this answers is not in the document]
chenzizhao: Yap it can be. I set it as var initially because I wasn't sure what `assert.async()` returns and forgot all about it .. Fixed

src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -0,0 +1,77 @@
...
+    const errorMsg = 'error message';
+
+    gapi.client.request.withArgs(REQUEST_BODY).returns(new Promise(() => {
```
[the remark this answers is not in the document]
chenzizhao: Done. I didn't know about `Promise.resolve` and `Promise.reject`. Thanks for pointing is out!

src/main/webapp/scripts/tests/comment-post-test.js
chenzizhao: Done

src/main/webapp/scripts/tests/comment-post-test.js
chenzizhao: Done!

src/main/webapp/scripts/tests/comment-post-test.js
chenzizhao: Yap Done!

src/main/webapp/scripts/post-comment.js
chenzizhao: Done

src/main/webapp/scripts/tests/comment-post-test.js
chenzizhao: Done

src/main/webapp/scripts/tests/comment-post-test.js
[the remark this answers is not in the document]
chenzizhao: Done. Mocking in Sinon seems tricky so I created my own fake classes. Now no need for loading the client :D

src/main/webapp/scripts/tests/comment-post-test.js
juanalonsocruz: I meant something more like 

```gapi.client.request.withArgs(createExpectedRequestBody(VIDEO_ID, COMMENT_TEXT)).returns(...)```

You can keep your new assertion as it is now but instead of REQUEST_BODY, it could be createExpectedRequestBody(...). Then when you read this unit test as a whole you see that when you call your method with video id and comment text, you successfully called the api with request body that contains the same video id and comment text.

chenzizhao: Oow my bad, I misunderstood the first time. It has been modified as suggested. I thought this will re-implement the logic from the main code, but I can definitely see your point of how it improves test readability.

PR #14, opened 2020-07-16
OAuth2 and post comment in the front end

chenzizhao:
This function allows user to authorize the site to post a comment on their behalf (the UI and other chores around #13 ). The comment text can be dynamically changed in the textarea in index.html. The target video is hard coded in view.js for now, and can be changed later. 

* Model: comments.js, View: view.js, Controller: controller_auth.js

#### If the comment is successfully posted: 
image
#### Otherwise, the returned error message is displayed. 
image

If the "Sign out" and "Post a comment" buttons do not load, it is likely that the API key and clientId are missing in controller_auth.js

### DIFF
--- a/src/main/webapp/scripts/controller-auth.js
+++ b/src/main/webapp/scripts/controller-auth.js
@@ -62,4 +62,4 @@ function handleCommentClick(event) {
 handleClientLoad();
 
 // For testing purpose
-export { initClient, handleAuthClick, handleSignoutClick, handleCommentClick };
+export { initClient };
--- a/src/main/webapp/scripts/controller-auth.js
+++ b/src/main/webapp/scripts/controller-auth.js
@@ -1,5 +1,5 @@
 import { postVideoComment } from '/scripts/post-comment.js';
-import { videoId, commentElement, failedInitCallback, successfulApiCallback, failedApiCallback, updateSigninStatus } from '/scripts/view.js';
+import { videoId, postCommentElement, failedInitCallback, successfulApiCallback, failedApiCallback, updateSigninStatus } from '/scripts/view.js';
 
 /** API key provied by GCP to authenticate the client */
 const apiKey = '';
--- a/src/main/webapp/scripts/controller-auth.js
+++ b/src/main/webapp/scripts/controller-auth.js
@@ -1,15 +1,24 @@
 import { postVideoComment } from '/scripts/post-comment.js';
-import { videoId, commentElement, failedInit, successfulApiCallback, failedApiCallback, updateSigninStatus } from '/scripts/view.js';
+import { videoId, commentElement, failedInitCallback, successfulApiCallback, failedApiCallback, updateSigninStatus } from '/scripts/view.js';
 
+/** API key provied by GCP to authenticate the client */
 const apiKey = '';
+/** client ID provied by GCP to identify the client */
 const clientId = '';
+/** discovery docs used to identify API endpoints */
 const discoveryDocs = ['https://www.googleapis.com/discovery/v1/apis/youtube/v3/rest'];
+/** OAuth2 scope required to post a comment */
 const scope = 'https://www.googleapis.com/auth/youtube.force-ssl';
 
+/** Load the client from Google API oauth2.0 JS library */
 function handleClientLoad() {
   gapi.load('client:auth2', initClient);
 }
 
+/** 
+ * Initialize the client with appropiate fields.
+ * And listen for changed in the signin status.
+ */
 function initClient() {
   gapi.client.init({
     apiKey,
@@ -26,17 +35,23 @@ function initClient() {
       updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get(),
         handleAuthClick, handleSignoutClick, handleCommentClick);
     })
-    .catch(failedInit);
+    .catch(failedInitCallback);
 }
 
+/** 
+ * Handler that initiates the authorization flow with a popup.
+ * The user only needs to authorize the site once.
+ */
 function handleAuthClick(event) {
   gapi.auth2.getAuthInstance().signIn();
 }
 
+/** Handler that quietly signs the user out, but does not revoke the authorization */
 function handleSignoutClick(event) {
   gapi.auth2.getAuthInstance().signOut();
 }
 
+/** Handler that posts the comment and processes the result */
 function handleCommentClick(event) {
   postVideoComment(videoId, postCommentElement.value)
     .then(successfulApiCallback)
--- a/src/main/webapp/scripts/controller-auth.js
+++ b/src/main/webapp/scripts/controller-auth.js
@@ -1,0 +1,47 @@
+import { postVideoComment } from '/scripts/post-comment.js';
+import { videoId, commentElement, failedInit, successfulApiCallback, failedApiCallback, updateSigninStatus } from '/scripts/view.js';
+
+const apiKey = '';
+const clientId = '';
+const discoveryDocs = ['https://www.googleapis.com/discovery/v1/apis/youtube/v3/rest'];
+const scope = 'https://www.googleapis.com/auth/youtube.force-ssl';
+
+function handleClientLoad() {
+  gapi.load('client:auth2', initClient);
+}
+
+function initClient() {
+  gapi.client.init({
+    apiKey,
+    discoveryDocs,
+    clientId,
+    scope
+  })
+    .then(function () {
+      // Listen for sign-in state changes.
+      gapi.auth2.getAuthInstance().isSignedIn.listen((isSignedIn) =>
+        updateSigninStatus(isSignedIn, handleAuthClick, handleSignoutClick, handleCommentClick));
+
+      // Handle the initial sign-in state.
+      updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get(),
+        handleAuthClick, handleSignoutClick, handleCommentClick);
+    })
+    .catch(failedInit);
+}
+
+function handleAuthClick(event) {
+  gapi.auth2.getAuthInstance().signIn();
+}
+
+function handleSignoutClick(event) {
+  gapi.auth2.getAuthInstance().signOut();
+}
+
+function handleCommentClick(event) {
+  postVideoComment(videoId, postCommentElement.value)
+    .then(successfulApiCallback)
+    .catch(failedApiCallback);
+}
+
+// Entry point
+handleClientLoad();
--- a/src/main/webapp/scripts/controller-auth.js
+++ b/src/main/webapp/scripts/controller-auth.js
@@ -20,7 +20,7 @@ function handleClientLoad() {
  * And listen for changed in the signin status.
  */
 function initClient() {
-  gapi.client.init({
+  return gapi.client.init({
     apiKey,
     discoveryDocs,
     clientId,
@@ -53,10 +53,13 @@ function handleSignoutClick(event) {
 
 /** Handler that posts the comment and processes the result */
 function handleCommentClick(event) {
-  postVideoComment(videoId, postCommentElement.value)
+  return postVideoComment(videoId, postCommentElement.value)
     .then(successfulApiCallback)
     .catch(failedApiCallback);
 }
 
 // Entry point
 handleClientLoad();
+
+// For testing purpose
+export { initClient, handleAuthClick, handleSignoutClick, handleCommentClick };
--- a/src/main/webapp/scripts/post-comment.js
+++ b/src/main/webapp/scripts/post-comment.js
@@ -20,7 +20,7 @@ function postVideoComment(videoId, commentText) {
  * @private
  * @param {String} videoId 
  * @param {String} commentContent 
- * @returns {json} request body for a POST comment request by video ID
+ * @returns {Object} request body for a POST comment request by video ID
  */
 function buildRequestBody(videoId, commentContent) {
     return {
--- a/src/main/webapp/scripts/post-comment.js
+++ b/src/main/webapp/scripts/post-comment.js
@@ -1,3 +1,11 @@
+/**
+ * Post a video comment by making a POST request to YouTube Data API.
+ * @exports
+ * @param {String} videoId 
+ * @param {String} commentText 
+ * @returns {String} empty string if successful post
+ * @throws {Error} with error message if failed post
+ */
 function postVideoComment(videoId, commentText) {
     return new Promise((resolve, reject) => {
         gapi.client.request(buildRequestBody(videoId, commentText))
@@ -7,6 +15,13 @@ function postVideoComment(videoId, commentText) {
     });
 }
 
+/**
+ * Build a request body expected by the API.
+ * @private
+ * @param {String} videoId 
+ * @param {String} commentContent 
+ * @returns {json} request body for a POST comment request by video ID
+ */
 function buildRequestBody(videoId, commentContent) {
     return {
         'path': '/youtube/v3/commentThreads?part=snippet',
--- a/src/main/webapp/scripts/tests/controller-auth-test.js
+++ b/src/main/webapp/scripts/tests/controller-auth-test.js
@@ -11,8 +11,7 @@ QUnit.module('Authorization controller', {
 });
 
 QUnit.test('When init fails, displays the authorization button and the error message', async function (assert) {
-  const errorMsg = 'error message';
-  sinon.stub(gapi.client, 'init').returns(Promise.reject(new Error(errorMsg)));
+  sinon.stub(gapi.client, 'init').returns(Promise.reject(new Error()));
 
   await controller.initClient();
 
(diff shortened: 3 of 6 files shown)


src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,46 @@
+import { postVideoComment } from '/scripts/post-comment.js';
+import { videoId, commentElement } from '/scripts/view.js';
```
FuzzyTree, line 2: combine this import with line 3

src/main/webapp/scripts/view.js
```diff
@@ -0,0 +1,41 @@
...
+function successfulApiCallback(data) {
+    updateDom('successful', resultContainer);
+}
+
+function failedApiCallback(err) {
+    console.error('Execution error ', err);
+    updateDom(err, resultContainer);
+}
+
+export { videoId, commentElement, resultContainer };
```
FuzzyTree, line 40: you can combine the two exports

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,46 @@
...
+  gapi.auth2.getAuthInstance().signIn();
+}
+
+function handleSignoutClick(event) {
+  gapi.auth2.getAuthInstance().signOut();
+}
+
+function handleCommentClick(event) {
+  postVideoComment(videoId, commentElement.value)
+    .then(data => successfulApiCallback(data))
```
FuzzyTree, line 41: Because `successfulApiCallback` takes the same argument as the promise result, you can simplify this line to `then(successFulApiCallBack)`

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,46 @@
...
+    .catch(error => failedApiCallback(error));
```
FuzzyTree, line 42: Similarly `catch(failedApiCallback)`

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,46 @@
...
+    scope: scopes
+  })
+    .then(function () {
+      // Listen for sign-in state changes.
+      gapi.auth2.getAuthInstance().isSignedIn.listen(updateSigninStatus);
+
+      // Handle the initial sign-in state.
+      updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get(),
+        handleAuthClick, handleSignoutClick, handleCommentClick);
+    });
```
FuzzyTree, line 28: may want to add an error handler in case init fails

src/main/webapp/scripts/controller-auth.js
chenzizhao: Done. Wasn't aware of this rule in the style guide but now I do. Thanks for pointing it out :)

src/main/webapp/scripts/view.js
chenzizhao: Done

src/main/webapp/scripts/controller-auth.js
chenzizhao: Done. This looks much nicer

src/main/webapp/scripts/controller-auth.js
chenzizhao: Done

src/main/webapp/scripts/controller-auth.js
chenzizhao: Done

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,46 @@
...
+const discoveryDocs = ['https://www.googleapis.com/discovery/v1/apis/youtube/v3/rest'];
+const scopes = 'https://www.googleapis.com/auth/youtube.force-ssl';
+
+function handleClientLoad() {
+  gapi.load('client:auth2', initClient);
+}
+
+function initClient() {
+  gapi.client.init({
+    apiKey: apiKey,
```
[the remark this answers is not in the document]
chenzizhao: Done. How neat!

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,61 @@
+import { postVideoComment } from '/scripts/post-comment.js';
+import { videoId, commentElement, failedInitCallback, successfulApiCallback, failedApiCallback, updateSigninStatus } from '/scripts/view.js';
+
+/** API key provied by GCP to authenticate the client*/
+const apiKey = '';
+/** client ID provied by GCP to identify the client*/
```
FuzzyTree, line 6: Nit: please leave a space before `*/` here and elsewhere i.e. `client */`

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,61 @@
...
+      updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get(),
+        handleAuthClick, handleSignoutClick, handleCommentClick);
+    })
+    .catch(failedInitCallback);
+}
+
+/** 
+ * Handler that initiates the authorization flow with a popup.
+ * The user only needs to authorize the site once.
+*/
```
FuzzyTree, line 43: nit: this should line up i.e. 

```
/** 
 * Handler that initiates the authorization flow with a popup.
 * The user only needs to authorize the site once.
 */
```

please also fix elsewhere

src/main/webapp/scripts/controller-auth.js
chenzizhao: Yap done

src/main/webapp/scripts/controller-auth.js
chenzizhao: Done thanks for pointing it out :D

src/main/webapp/scripts/view.js
```diff
@@ -0,0 +1,79 @@
+/** Botton that triggers authorization/sign out flow */
```
juanalonsocruz, line 1: Whole PR: You added a controller and view logic. You need to add tests to make sure it works :). Mock the responses from the API and test that your controller does what it does and sets the view correctly. Mocking the listen probably won't work since its probably doing some fancy stuff in it. So you probably are fine with just calling the methods directly and making sure the view is changed.

src/main/webapp/scripts/view.js
```diff
@@ -0,0 +1,79 @@
...
+const authorizeButton = document.getElementById('authorize-button');
+
+/** Hardcoded test video Id that the comment will be posted to*/
+const videoId = 'gjKpAAezWQA';
+
+/** Textbox that allows user to input comments*/
+const commentElement = document.getElementById('comment-text');
+
+/** Container that holds the success/error message*/
+const resultContainer = document.getElementById('result-text');
```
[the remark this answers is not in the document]
chenzizhao: Renamed `commentElement` to `postCommentElement` and `resultContainer` to `postResultContainer`

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,61 @@
...
+function initClient() {
+  gapi.client.init({
+    apiKey,
+    discoveryDocs,
+    clientId,
+    scope
+  })
+    .then(function () {
+      // Listen for sign-in state changes.
+      gapi.auth2.getAuthInstance().isSignedIn.listen(updateSigninStatus);
```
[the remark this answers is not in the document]
chenzizhao: You are right! This resolved a UI bug that only appears when the user authorize for the first time, I was able to fix the bug by hard refreshing at that time so I didn't dig into it .. but I think this is root cause. Thank you!

src/main/webapp/scripts/view.js
chenzizhao: Ok that makes sense. Just to confirm are you referring to some kind of integration test involving DOM instead of unit testing a single function like handleCommentClick()? (your comment on Jennifer's PR about quit-dom will be helpful :D)

src/main/webapp/scripts/tests/controller-auth-test.js
```diff
@@ -0,0 +1,106 @@
...
+
+QUnit.test('AWAIT When init fails, displays the authorization button and the error message', async function (assert) {
+  const errorMsg = 'error message';
+  sinon.stub(gapi.client, 'init').returns(Promise.reject(new Error(errorMsg)));
+
+  await controller.initClient();
+
+  assert.dom('#result-text').hasProperty('innerText', 'Failed to initialize client');
+  gapi.client.init.restore();
+});
```
chenzizhao, line 38: Aims to test the same function but with different method

chenzizhao:
Tests have been added in src/main/webapp/scripts/tests/controller-auth-test.js 

The last two tests are more like integration tests as they asserts the changes in DOM caused by clicking a button. In those two tests I stubbed `gapi.client` which results in `model.postVideoComment` being tested again besides in its stand alone unit tests in /scripts/tests/post-comment-test.js  I considered a better way of doing this is to stub the response of `model.postVideoComment` but I failed to stub a custom module (https://railsware.com/blog/mocking-es6-module-import-without-dependency-injection/ and https://sinonjs.org/how-to/stub-dependency/) with the following error. 

```
TypeError: ES Modules cannot be stubbed
    at Function.stub (https://sinonjs.org/releases/sinon-9.0.0.js:3063:15)
    at Sandbox.stub (https://sinonjs.org/releases/sinon-9.0.0.js:2625:37)
    at Object.<anonymous> (http://localhost:8070/scripts/tests/controller-auth-test.js:51:9)
```

src/main/webapp/scripts/controller-auth.js
```diff
@@ -0,0 +1,65 @@
...
+  return postVideoComment(videoId, postCommentElement.value)
+    .then(successfulApiCallback)
+    .catch(failedApiCallback);
+}
+
+// Entry point
+handleClientLoad();
+
+// For testing purpose
+export { initClient, handleAuthClick, handleSignoutClick, handleCommentClick };
```
juanalonsocruz, line 65: Is there a way to test this module without exposing everything in this file? When you export in a file you are making an interface telling others that they can use these when that might be unintended.

src/main/webapp/scripts/tests/controller-auth-test.js
```diff
@@ -0,0 +1,93 @@
+import * as view from '/scripts/view.js';
+import * as controller from '/scripts/controller-auth.js';
+
+// Test (controller-auth.js and) view.js
+QUnit.module('Authorization controller', {
+  beforeEach: () => {
+    gapi = new CustomGapi();
+  },
+  afterEach: () => {
+    view.updateSigninStatus(false, null, null, null);
```
juanalonsocruz, line 10: Is this needed? Doesn't the controller initialize the view in the beginning?

src/main/webapp/scripts/tests/controller-auth-test.js
```diff
@@ -0,0 +1,93 @@
...
+    document.getElementById('result-text').innerText = '';
```
juanalonsocruz, line 11: Is this needed? If it is, maybe put in a beforeEach instead. It makes it clear that the page is reset before starting a test.

src/main/webapp/scripts/tests/controller-auth-test.js
```diff
@@ -0,0 +1,93 @@
...
+    sinon.restore();
+  }
+});
+
+QUnit.test('When init fails, displays the authorization button and the error message', async function (assert) {
+  const errorMsg = 'error message';
```
juanalonsocruz, line 17: Nit: Seems like this message is irrelevant to the test. Below, you could just do `new Error()`

src/main/webapp/scripts/tests/controller-auth-test.js
```diff
@@ -0,0 +1,93 @@
...
+  await controller.initClient();
+
+  assert.dom('#result-text').hasProperty('innerText', 'Failed to initialize client');
+});
+
+QUnit.test('When signed in, there should be sign-out and comment buttons', function (assert) {
+  const fakeSignOut = sinon.fake();
+  const fakeHandleCommentClick = sinon.fake();
+
+  view.updateSigninStatus(true, () => { }, fakeSignOut, fakeHandleCommentClick);
```
juanalonsocruz, line 29: Is it possible to test this by mocking the gapi client init to immediately return a resolved promised and gapi.auth2.getAuthInstance().isSignedIn.get() to return true so it could call updateSigninStatus through its natural flow?

juanalonsocruz: Similar comment for every test that directly uses the view.

src/main/webapp/scripts/controller-auth.js
chenzizhao: handleClientLoad / initClick are already somehow exposed because they are run onload. 

And I agree about encapsulating {handleAuthClick, handleSignoutClick, handleCommentClick} -- actually I have tried the natural way of invoking handleCommentClick like this

```
postCommentButton.click()

assert.dom()...
```

but `postCommentButton.click()` isn't an aync function so `async/await` doesn't work as intended in the tests (setTimeOut works most of the time but flaky).

juanalonsocruz: Sorry I meant if there is a way to only expose the "public" methods only. 

For the click issue, idk if what I'm about to say is a hack. How about, using a queryselector to get the button, then somehow retrieve the function/callback that is called when the button is clicked, then override the click function to return a promise that calls the function/callback you retrieved, and then you can do .click() and await? lol idk if that is how this works haha.

src/main/webapp/scripts/tests/controller-auth-test.js
```diff
@@ -0,0 +1,111 @@
...
+QUnit.test('When signed in, there should be sign-out and comment buttons', async function (assert) {
+  sinon.stub(gapi.client, 'init').returns(Promise.resolve());
+  sinon.stub(gapi.auth2, 'getAuthInstance').returns(
+    {
+      'isSignedIn':
+      {
+        'get': () => { return true; },
+        'listen': () => { },
+      }
+    });
```
juanalonsocruz, line 32: Put this into a helper method that accepts the isSignedIn boolean as a parameter and call it in your tests.

src/main/webapp/scripts/tests/controller-auth-test.js
chenzizhao: Removed. Thanks!

src/main/webapp/scripts/tests/controller-auth-test.js
chenzizhao: Done

src/main/webapp/scripts/tests/controller-auth-test.js
chenzizhao: This error message actually shows up in DOM and it changes when deployed (video not found, id cannot be empty etc)

src/main/webapp/scripts/tests/controller-auth-test.js
chenzizhao: Done.

src/main/webapp/scripts/tests/controller-auth-test.js
chenzizhao: Done.

src/main/webapp/scripts/tests/controller-auth-test.js
chenzizhao: oow wait I messed up you are right haha

src/main/webapp/scripts/controller-auth.js
chenzizhao: Yap it worked! The only thing is that `await button.click()` doesn't work as expected so I changed it to `await button.onclick()` wasn't so sure about the difference ...

chenzizhao:
Also to followup on sinon unable to stub a custom module, @nif33 found this thread which explains how `import` gets in the way of stubbing.  https://github.com/sinonjs/sinon/issues/1711#issuecomment-369220753

src/main/webapp/scripts/view.js
nif33: wow it's cool that github notified me that I was mentioned!
I asked a question about this before and learned that testing controller functions (eg. get some data, do an api call, get response, display response) is considered a unit test and testing the helper methods is not always necessary (see comment on PR #12)

src/main/webapp/scripts/tests/controller-auth-test.js
```diff
@@ -0,0 +1,119 @@
...
+  sinon.stub(gapi.client, 'request').returns(Promise.reject(new CustomError(errorMsg)));
+
+  await controller.initClient();
+
+  const postCommentButton = document.getElementById('comment-button');
+  const temp = postCommentButton.onclick;
+  postCommentButton.onclick = () => {
+    return Promise.resolve(temp());
+  };
+  await postCommentButton.onclick();
```
juanalonsocruz, line 54: LOL! It worked.

src/main/webapp/scripts/post-comment.js
```diff
@@ -7,6 +15,13 @@ function postVideoComment(videoId, commentText) {
     });
 }
 
+/**
+ * Build a request body expected by the API.
+ * @private
+ * @param {String} videoId 
+ * @param {String} commentContent 
+ * @returns {json} request body for a POST comment request by video ID
```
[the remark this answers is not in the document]
chenzizhao: Done!

PR #16, opened 2020-07-17
add functions for region selection

nif33:
adds two functions:
-function for fetching regions list from API
-function for adding a new option to a selection list in HTML

these functions are helper functions, they will be tested within the outer function that calls them


src/main/webapp/scripts/region-selector.js
```diff
@@ -0,0 +1,13 @@
+export { getRegions }
```
chenzizhao, line 1: nit: Juan: export is usually placed at the end of the file

src/main/webapp/scripts/region-selector.js
```diff
@@ -0,0 +1,13 @@
...
+function getRegions() {
+  return gapi.client.youtube.i18nRegions.list({
+    "part": [
+      "snippet"
+    ],
+    "hl": "es_MX"
+  })
+  .then(function(response) {
+    return response.result;
+  });
```
chenzizhao, line 12: Do you catch the error when you call getRegions()?

src/main/webapp/scripts/script.js
```diff
@@ -8,3 +8,11 @@ function updateDom(text, containerId) {
 function getDomValue(domId) {
   return document.getElementById(domId).value;
 }
+
+function makeSelectList(value, text, selectId) {
```
chenzizhao, line 12: The function name "makeSelectList" seems to me that it returns an entire select element with all the options set up, while what it actually does is append one option to the select element. What do you think?

src/main/webapp/scripts/region-selector.js
```diff
@@ -0,0 +1,17 @@
+function getRegions() {
+  return gapi.client.youtube.i18nRegions.list({
+    "part": [
+      "snippet"
+    ],
+    "hl": "es_MX"
```
chenzizhao, line 6: nit: single quote lol

src/main/webapp/scripts/region-selector.js
```diff
@@ -0,0 +1,17 @@
+function getRegions() {
+  return gapi.client.youtube.i18nRegions.list({
+    'part': [
+      'snippet'
+    ],
+    'hl': 'es_MX'
```
chenzizhao, line 6: nit: maybe add a comment to explain what this field is about?

juanalonsocruz: Agreed. Can't tell what this is either.

src/main/webapp/scripts/region-selector.js
```diff
@@ -0,0 +1,17 @@
+function getRegions() {
```
juanalonsocruz, line 1: You are currently adding "dead code" (code that won't run). When adding functionality, you should also add its usage (unless you creating an API method).  In this case, it is showing the list of regions in the UI. Please add this. Also add test for that.

src/main/webapp/scripts/region-selector.js
```diff
@@ -0,0 +1,17 @@
...
+      'snippet'
+    ],
+    'hl': 'es_MX'
+  })
+  .then((response) => {
+    return response.result;
+  })
+  .catch((error) => {
+    console.error('An error occurred getting the region list:', error);
+    throw error;
```
juanalonsocruz, line 13: doing a console.error and throwing the error will make it have two logs in the console. If the error is informative enough then you don't need to catch. For future PR, maybe you want to catch it to have a different behavior on the UI? Like not letting them submit anything.

src/main/webapp/scripts/region-selector.js
nif33: I realized I didn't even need it 😅 I don't remember how it got there, probably from some documentation somewhere...

src/main/webapp/tests.html
```diff
@@ -7,16 +7,21 @@
...
   <script src="https://sinonjs.org/releases/sinon-9.0.0.js"></script>
+  <script src="https://apis.google.com/js/api.js"></script>
   <script src="scripts/script.js"></script>
   <script src="scripts/perspective.js"></script>
   <script src="scripts/perspective-api-service.js"></script>
   <script src="scripts/classes/comment-class.js"></script>
   <script src="scripts/tests/perspective-test.js"></script>
-  <script src="https://apis.google.com/js/api.js"></script>
+  <script type="module" src="scripts/tests/popular-test.js"></script>
   <script defer type="module" src="scripts/tests/comment-post-test.js"></script>
```
FuzzyTree, line 25: is `defer` necessary?

src/main/webapp/scripts/script.js
```diff
@@ -1,4 +1,24 @@
 const API_KEY = '';
+const YOUTUBE_DISCOVERY = 'https://www.googleapis.com/discovery/v1/apis/youtube/v3/rest';
+
+function loadApi() {
+  return new Promise((resolve, reject) => {
+    gapi.load('client', function() {
```
FuzzyTree, line 6: Nit: please convert all anonymous (unnamed) functions to arrow functions for consistency.

`gapi.load('client', () => {`

src/main/webapp/scripts/tests/popular-test.js
```diff
@@ -0,0 +1,18 @@
...
+  items: [
+    { snippet: { gl: 'EG', name: 'Example'}}
+  ]
+}
+
+QUnit.test('region menu display',  function(assert) {
+  const done = assert.async();
+  const getRegions = sinon.fake.returns(
+    Promise.resolve(mockGetRegionsResponse)
+  );
```
FuzzyTree, line 13: This will not work because `getRegions` is in a different scope from this function. One way to mock getRegions is to import its reference from popular.js:

```
import * as popular from '../popular.js';
.
.
.
let { addRegionOptions, getRegions } = popular;
getRegions = sinon.fake ...
```

You'll also need to export `getRegions` from popular.js

src/main/webapp/tests.html
```diff
@@ -7,16 +7,21 @@
   <link rel="stylesheet" href="https://code.jquery.com/qunit/qunit-2.10.0.css">
 </head>
 <body>
+    <select id="region-select"></select>
   <div id="qunit"></div>
-  <div id="qunit-fixture"></div>
+  <div id="qunit-fixture">
+
+  </div>
```
FuzzyTree, line 14: Is this change intentional?

nif33: going to put the select dropdown menu here once it's working (it's not here right now because this section hides the element and I want to see the box when debugging)

src/main/webapp/tests.html
nif33: I don't think I need it for mine, @chenzizhao why was it added comment-post-test.js?

chenzizhao: ooh sorry it works without defer here my bad, please remove it

src/main/webapp/scripts/popular.js
```diff
@@ -0,0 +1,12 @@
+import { getRegions } from './region-selector.js';
+
+function addRegionOptions() {
```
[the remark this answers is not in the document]
nif33: I don't have access to that link but assuming it is the same info as the public style guide I have added some JSDoc

src/main/webapp/scripts/tests/popular-test.js
nif33: It's still not working :/ did I interpret wrong? I am still getting no error message but it is using the real function instead of the fake. I also tried exporting getRegions in popular-test and `import { getRegions } from './tests/popular-test'` in the popular.js file but that caused an error because the functions have the same name. How can I specify when I want to use the getRegions in the original file vs the getRegions fake from the test file...

FuzzyTree: Sorry my instructions were not right. Your export in popular.js should be a default export:

`export default {addRegionOptions, getRegions}`

and the import in this file should be 

`import popular from '../popular.js';`

you should use sinon.replace to mock getRegions

```
QUnit.test('region menu display',  function(assert) {
  const done = assert.async();
  const getRegions = sinon.fake.returns(
    Promise.resolve(mockGetRegionsResponse)
  );
  sinon.replace(popular, 'getRegions', getRegions);
  popular.addRegionOptions().then(() => {
    assert.dom('option').hasText('Example');
    done();
  });
});
```

FuzzyTree: Sorry, my instructions were not right. This works on my machine:
```
//popular.js
import { getRegions } from './region-selector.js';
let popular;

function addRegionOptions() {
  return loadApi().then(() => popular.getRegions()) // call popular.getRegions() here
  .then((response) => {
    for(const region of response.items) {
      addOptionToSelectList(region.snippet.gl, region.snippet.name, 'region-select');
    }
  });
}   
  
export default popular = {addRegionOptions, getRegions} 
```

```import popular from '../popular.js';

const mockGetRegionsResponse = {
  items: [
    { snippet: { gl: 'EG', name: 'Example'}}
  ]
}   
      
QUnit.test('region menu display',  function(assert) {
  const done = assert.async();
  const getRegions = sinon.fake.returns(
    Promise.resolve(mockGetRegionsResponse)
  );
  sinon.replace(popular, 'getRegions', getRegions);
  popular.addRegionOptions().then(() => {
    assert.dom('option').hasText('Example');
    done();
  });
});```

FuzzyTree: `let popular` is necessary because ES6 imports are readonly so you have to use `let` to be able to overwrite the import.

Normally ES6 code is transpiled to ES5 (to support older browsers) where the original instructions should work.

chenzizhao: Sorry to cut in -- what's the rationale behind exporting the dependency `export default popular = {addRegionOptions, getRegions}` in popular.js, rather than importing `getRegions` from `./region-selector.js` in popular-test.js?

FuzzyTree: Exporting the `getRegions` dependency from popular.js is needed order to get the same reference to `getRegions` that is being used in `popular.js`.

Note that in `popular.js` we're calling `loadApi().then(popular.getRegions)` instead of `loadApi().then(getRegions)` to ensure that the same reference as the export is used.

See the stackoverflow link above for more details.

chenzizhao: Thank you! I wasn't clear on the reference of function previously but now I get it.

PR #17, opened 2020-07-17
add functions that display menu and header

nif33:
-add function that fills out the header tag
-add function that adds menu items
I made these functions that build HTML because:
-when changes need to be made (such as setting our project name) it does not need to be done on every page
-if we need to add items to the menu it can be done quickly through making a new variable in MENU_ITEMS
Here is an image of the implemented interface:

image


src/main/webapp/scripts/script.js
```diff
@@ -8,3 +18,41 @@ function updateDom(text, containerId) {
 function getDomValue(domId) {
   return document.getElementById(domId).value;
 }
+
+function MenuItem (name, link, icon) {
```
ajpaon, line 22: It will be clearer that is meant to be used with `new` if you make this a `class`

I think it would also be reasonable to do away with this type altogether, and use raw objects for menu items, as in

```
const MENU_ITEMS = [
  {name: 'Trending', link: 'popular.html', icon: 'whatshot'},
  //...
}
```

src/main/webapp/scripts/script.js
```diff
@@ -1,5 +1,15 @@
...
 
+const TITLE_TEXT = 'Capstone Project' //TODO: change to actual name
+const MENU_ITEMS = [
+  new MenuItem('Trending', 'popular.html', 'whatshot'),
+  new MenuItem('Positive', 'positive.html', 'grade'),
+  new MenuItem('Check', 'index.html', 'check_circle')
+];
+
+createHeader();
+createMenu();
```
ajpaon, line 11: any thoughts about how to unit test these functions? :)

src/main/webapp/scripts/script.js
nif33: I went with the raw objects option because I didn't want to make a class -> put it in a new file -> load that script or have a class somewhere in the middle of all these constants and functions and whatnot

src/main/webapp/scripts/script.js
nif33: should I be testing DOM manipulations...? There is not much logic going on in these functions since they just create a bunch of HTML. I googled how to test dom manipulation and found this stack overflow answer: https://stackoverflow.com/questions/18852450/javascript-unit-testing-dom-manipulation
Some ideas I had were:
- I could test that getElementById did not return undefined, but it already throws an error if another element tries to append to the undefined (i could make it throw a better error to tell the caller that elements are missing from the HTML?)
- I could test that the correct elements were added to the html, but that is based on functions like appendChild that are built-in
- I could fake the document.method() calls and test that the for loop in createMenu() iterated through the expected items
Could I get more guidance on which of the above/what else should be tested for in this scenario

src/main/webapp/scripts/tests/view-test.js
```diff
@@ -0,0 +1,9 @@
+QUnit.test('createHeader', function(assert) {
+  createHeader();
+  assert.dom('h3').hasText(TITLE_TEXT);
+});
+
+QUnit.test('createMenu', function(assert) {
+  createMenu();
+  assert.dom('i').exists({ count: 3});
+});
```
juanalonsocruz, line 9: It is not obvious that the 'h3' and 'i' are for the header and menu. Could you instead add id attributes and query by those here instead? Easier to read if it said "assert that the header has TITLE_TEXT" and "assert the menu has 3 items".

src/main/webapp/style.css
```diff
@@ -0,0 +1,47 @@
...
+#menu {
+  width: 5rem;
+}
+
+a {
+  color: black;
+}
+a:hover {
+  color: black;
+}
```
chenzizhao, line 41: Is it overwriting something?

nif33: links are by default blue, but I did not need to change hover too so i have removed that

src/main/webapp/scripts/tests/view-test.js
nif33: changed it so that it queries by the bootstrap classes that the tags already had (navbar-brand and nav-item) which I think are descriptive enough to not have to add another class (assert that the navigation bar's brand is TITLE_TEXT and assert there are 3 navigation items)

PR #18, opened 2020-07-20
Returns empty String if no captions are found.

kayefung:
Small change to ```captions.java``` to return an empty String instead of null if no captions are found for a video. Also updated tests to reflect the change.

### DIFF
--- a/src/main/java/com/google/sps/servlets/Caption.java
+++ b/src/main/java/com/google/sps/servlets/Caption.java
@@ -29,7 +29,7 @@ public class Caption {
         continue;
       }
     }
-    return null;
+    return "";
   }
 
   private String parseXmlFromStream(InputStream inputStream)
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -29,7 +29,7 @@ public String getCaptionFromId(String videoId) {
         continue;
       }
     }
-    return null;
+    return "";
```
juanalonsocruz, line 32: Off topic: Does captionService.fetchStream(...) throw an error if the video id is not valid? I would find it out if this method return an empty string for an invalid video because it lets the caller think that its a valid video without captions.

chenzizhao: fetchStream throws a FileNotFound error if the video Id is invalid or the caption does not exist. So there isn't a way to differentiate that from this perspective.

kayefung: ^^ building off this, CommentService throws an IllegalArgumentException if the video can't be found and when SentimentServlet catches that it'll send an error to the client and stop the servlet from running

PR #19, opened 2020-07-20
Use SentimentServlet to retrieve video data and analyze sentiment.

kayefung:
When a user inputs a link to a YT video and click the analyze button the SentimentServlet.java will call CommentService.java and Caption.java to get the comments and captions of the inputted video. The text will get analyzed in calculateSentimentScore(). The score can be retrieved with a GET request (coming in an future PR)

I also mentioned in the pod sync that I'm having problems making multiple calls to the NLP API, for now we're going to put the captions and comments together into one String and analyze its sentiment as a whole.

### DIFF
--- a/src/main/java/com/google/sps/servlets/SentimentServlet.java
+++ b/src/main/java/com/google/sps/servlets/SentimentServlet.java
@@ -57,9 +57,10 @@ public class SentimentServlet extends HttpServlet {
     String captions = captionService.getCaptionFromId(videoId);
 
     if (comments.isEmpty() && captions.isEmpty()) {
-      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, NO_DATA_ERROR);
-
-      VideoAnalysis videoAnalysis = new VideoAnalysis(null, false);
+      VideoAnalysis videoAnalysis = new VideoAnalysis.Builder()
+          .setScore(null)
+          .setDataAvailable(false)
+          .build();
       String json = new Gson().toJson(videoAnalysis);
 
       response.setContentType("application/json;");
@@ -79,7 +80,10 @@ public class SentimentServlet extends HttpServlet {
       return;
     }
 
-    VideoAnalysis videoAnalysis = new VideoAnalysis(score, true);
+    VideoAnalysis videoAnalysis = new VideoAnalysis.Builder()
+        .setScore(score)
+        .setDataAvailable(true)
+        .build();
     String json = new Gson().toJson(videoAnalysis);
 
     response.setContentType("application/json;");
--- a/src/main/java/com/google/sps/servlets/SentimentServlet.java
+++ b/src/main/java/com/google/sps/servlets/SentimentServlet.java
@@ -26,7 +26,6 @@ public class SentimentServlet extends HttpServlet {
   private Caption captionService;
   private LanguageServiceClient languageService;
   private Sentiment sentiment;
-  private VideoAnalysis videoAnalysis;
   private final String INVALID_INPUT_ERROR = "Video is private or does not exist.";
   private final String COMMENTS_FAILED_ERROR = "Comments could not be retrieved.";
   private final String NLP_API_ERROR = "Language service client failed.";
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -19,71 +22,79 @@
...
+    String comments = String.join(". ", commentsList);
+
+    String captions = captionService.getCaptionFromId(videoId);
 
-    if (comments.isEmpty()) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, EMPTY_STRING_ERROR);
+    if (comments.isEmpty() && captions.isEmpty()) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, NO_DATA_ERROR);
       return;
     }
```
juanalonsocruz, line 77: Instead of any error for no data, why not just tell the client that as part of the response (VideoAnalysis)? Maybe add another field to VideoAnalysis that conveys this information.

src/main/webapp/scripts/analyze.js
```diff
@@ -0,0 +1,7 @@
+function analyzeVideo() {
+  const videoId = document.getElementById("video-url").value.split('v=')[1];
+  fetch('/sentiment', {
+    method: 'POST',
+    body: videoId
+  });
+}
```
juanalonsocruz, line 7: Keep JS changes to a different PR unless it is required due to breaking/changing a previous feature..

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -53,97 +61,159 @@ public void setUp() throws IOException {
...
         .thenReturn(sentimentMock);
     when(sentimentMock.getScore()).thenReturn(SCORE);
     doNothing().when(languageServiceMock).close();
 
+    sentimentServlet.doPost(requestMock, responseSpy);
+
     sentimentServlet.doGet(requestMock, responseSpy);
 
-    Assert.assertEquals(stringWriter.toString(), "{\"commentScore\":0.05}\n");
+    Assert.assertEquals(stringWriter.toString(), "{\"score\":0.05}\n");
```
juanalonsocruz, line 82: Could you build the expected string inside the unit test using the SCORE variable. Its not obvious that 0.05 is the value of SCORE by just reading the unit test.

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -19,71 +22,79 @@
...
     languageService = LanguageServiceClient.create();
+    captionService = new Caption();
+    commentService = new CommentService();
   }
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String comments = "";
-    Float commentsScore = null;
+    String json = new Gson().toJson(videoAnalysis);
```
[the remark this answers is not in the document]
chenzizhao: Just to add on to the suggestion: I experienced some difficulty to add a body to a GET request (in JavaScript) and I opted for query string instead..

https://stackoverflow.com/questions/978061/http-get-with-request-body#:~:text=Yes.,semantic%20meaning%20to%20the%20request.

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -50,100 +57,156 @@ public void setUp() throws IOException {
...
+    doNothing().when(languageServiceMock).close();
 
     sentimentServlet.doGet(requestMock, responseSpy);
 
+    String expected = "{\"score\":" + df.format(SCORE) + ",\"dataAvailable\":true}\n";
+
     verify(responseSpy).sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
-        "Language service client failed.");
+        "Comments could not be retrieved.");
+    Assert.assertEquals(expected, stringWriter.toString());
```
chenzizhao, line 190: Is it okay to send both the error and the value to the front end? For example, if I want to get the score in the front end, do I find it in fetch..then.. or fetch..catch.. ?

chenzizhao: nvm now I know you are reading the status and the message

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -19,71 +22,74 @@
...
+  private final String COMMENTS_FAILED_ERROR = "Comments could not be retrieved.";
   private final String NLP_API_ERROR = "Language service client failed.";
+  private final String NO_DATA_ERROR = "No comments or captions available to analyze.";
+  private final String READER_ERROR = "Reading video ID failed.";
 
-  public SentimentServlet() throws IOException {
+  public SentimentServlet() throws IOException, GeneralSecurityException {
     languageService = LanguageServiceClient.create();
+    captionService = new Caption();
+    commentService = new CommentService();
```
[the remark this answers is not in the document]
kayefung: It made injecting mocks during testing easier (if I put it in the constructor and avoid calling ```new SentimentServlet()``` then it won't replace the mocks with real instances of the classes). I definitely think it looks weird though lol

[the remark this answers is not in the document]
chenzizhao: ooook that makes sense, thanks!

PR #20, opened 2020-07-21
Display sentiment score to user.

kayefung:
Analyzes the user inputted video link and adds the sentiment score to the DOM. In the case that the fetch response is an error, the error message is added to the DOM instead.

### DIFF
--- a/src/main/webapp/scripts/analyze.js
+++ b/src/main/webapp/scripts/analyze.js
@@ -1,0 +1,44 @@
+function analyzeVideo() {
+  const videoId = document.getElementById('video-url').value.split('v=')[1];
+  const baseUrl = window.location.origin;
+  let url = new URL('/sentiment', baseUrl);
+  url.searchParams.append('video-id', videoId);
+
+  fetch(url)
+  .then(response => {
+    if (response.status >= 200 && response.status <= 299) {
+      return response.json();
+    } else if (response.status >= 400 && response.status <= 499) {
+      throw InvalidVideoException(response.statusText);
+    } else {
+      throw ServletFailureException(response.statusText);
+    }
+  })
+  .then(videoAnalysis => {
+    if (videoAnalysis.dataAvailable) {
+      updateSentimentContainer(videoAnalysis.score);
+    } else {
+      updateSentimentContainer("We couldn't analyze this video! Try again.")
+    }
+  })
+  .catch(error => {
+    updateSentimentContainer(error.message);
+  });
+}
+
+function updateSentimentContainer(text) {
+  const container = document.getElementById('sentiment-container');
+  container.innerText = text.toString();
+}
+
+function InvalidVideoException(message, metadata) {
+  const error = new Error(message);
+  error.metadata = metadata;
+  return error;
+}
+
+function ServletFailureException(message, metadata) {
+  const error = new Error(message);
+  error.metadata = metadata;
+  return error;
+}
--- a/src/main/webapp/scripts/analyze.js
+++ b/src/main/webapp/scripts/analyze.js
@@ -1,44 +1,23 @@
-function analyzeVideo() {
+async function analyzeVideo() {
   const videoId = document.getElementById('video-url').value.split('v=')[1];
-  const baseUrl = window.location.origin;
-  let url = new URL('/sentiment', baseUrl);
-  url.searchParams.append('video-id', videoId);
 
-  fetch(url)
+  return fetch(`/sentiment?video-id=${videoId}`)
   .then(response => {
     if (response.status >= 200 && response.status <= 299) {
       return response.json();
-    } else if (response.status >= 400 && response.status <= 499) {
-      throw InvalidVideoException(response.statusText);
     } else {
-      throw ServletFailureException(response.statusText);
+      throw new Error(response.statusText);
     }
   })
   .then(videoAnalysis => {
-    if (videoAnalysis.dataAvailable) {
-      updateSentimentContainer(videoAnalysis.score);
+    if (videoAnalysis.scoreAvailable) {
+      updateDom(videoAnalysis.score.toString(), 'sentiment-container');
     } else {
-      updateSentimentContainer("We couldn't analyze this video! Try again.")
+      updateDom('We couldn\'t analyze this video! Try again.',
+          'sentiment-container');
     }
   })
   .catch(error => {
-    updateSentimentContainer(error.message);
+    updateDom(error.message, 'sentiment-container');
   });
-}
-
-function updateSentimentContainer(text) {
-  const container = document.getElementById('sentiment-container');
-  container.innerText = text.toString();
-}
-
-function InvalidVideoException(message, metadata) {
-  const error = new Error(message);
-  error.metadata = metadata;
-  return error;
-}
-
-function ServletFailureException(message, metadata) {
-  const error = new Error(message);
-  error.metadata = metadata;
-  return error;
 }
--- a/src/main/webapp/scripts/analyze.js
+++ b/src/main/webapp/scripts/analyze.js
@@ -3,7 +3,7 @@ async function analyzeVideo() {
 
   return fetch(`/sentiment?video-id=${videoId}`)
   .then(response => {
-    if (response.status >= 200 && response.status <= 299) {
+    if (response.ok) {
       return response.json();
     } else {
       throw new Error(response.statusText);
@@ -13,7 +13,7 @@ async function analyzeVideo() {
     if (videoAnalysis.scoreAvailable) {
       updateDom(videoAnalysis.score.toString(), 'sentiment-container');
     } else {
-      updateDom('We couldn\'t analyze this video! Try again.',
+      updateDom('Video has no available captions or comments to analyze.',
           'sentiment-container');
     }
   })
--- a/src/main/webapp/scripts/tests/analyze-test.js
+++ b/src/main/webapp/scripts/tests/analyze-test.js
@@ -1,67 +1,57 @@
-const VIDEO_LINK = "https://www.youtube.com/watch?v=testvideoid";
+const VIDEO_LINK = 'https://www.youtube.com/watch?v=testvideoid';
 const SCORE = 0.05;
+let fetchStub = null;
 
-QUnit.module('Updating DOM after analyzing video');
+QUnit.module('Updating DOM after analyzing video', {
+  beforeEach: function() {
+    document.getElementById('sentiment-container').innerText = '';
+    fetchStub = sinon.stub(window, 'fetch');
+  },
+  afterEach: function() {
+    fetch.restore();
+  }
+});
+
 QUnit.testStart(() => {
   document.getElementById('video-url').innerText = VIDEO_LINK;
-})
+});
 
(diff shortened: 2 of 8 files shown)


src/main/webapp/scripts/analyze.js
```diff
@@ -0,0 +1,44 @@
...
+  url.searchParams.append('video-id', videoId);
+
+  fetch(url)
+  .then(response => {
+    if (response.status >= 200 && response.status <= 299) {
+      return response.json();
+    } else if (response.status >= 400 && response.status <= 499) {
+      throw InvalidVideoException(response.statusText);
+    } else {
+      throw ServletFailureException(response.statusText);
```
juanalonsocruz, line 14: Whoa didn't know fetch does not catch all HTTP errors. The http responses in the backend are descriptive so I believe it is ok to just do a generic JS error and log it. No need for the custom errors.
Found in https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
```
function handleErrors(response) {
    if (!response.ok) {
        throw Error(response.statusText);
    }
    return response;
}
fetch("http://httpstat.us/500")
    .then(handleErrors)
    .then(response => console.log("ok") )
    .catch(error => console.log(error) );
```

src/main/webapp/scripts/analyze.js
```diff
@@ -0,0 +1,44 @@
...
+  })
+  .then(videoAnalysis => {
+    if (videoAnalysis.dataAvailable) {
+      updateSentimentContainer(videoAnalysis.score);
+    } else {
+      updateSentimentContainer("We couldn't analyze this video! Try again.")
+    }
+  })
+  .catch(error => {
+    updateSentimentContainer(error.message);
```
juanalonsocruz, line 25: Could you also log into the console? `console.error(error)`

src/main/webapp/scripts/analyze.js
```diff
@@ -0,0 +1,44 @@
...
+      throw InvalidVideoException(response.statusText);
+    } else {
+      throw ServletFailureException(response.statusText);
+    }
+  })
+  .then(videoAnalysis => {
+    if (videoAnalysis.dataAvailable) {
+      updateSentimentContainer(videoAnalysis.score);
+    } else {
+      updateSentimentContainer("We couldn't analyze this video! Try again.")
```
juanalonsocruz, line 21: I believe the dataAvailable field is to tell the caller that there is no score. Maybe a different variable name would imply that better.

src/main/webapp/scripts/tests/analyze-test.js
```diff
@@ -0,0 +1,67 @@
...
+
+QUnit.test('DOM updates with sentiment score', function (assert) {
+  const fetchStub = sinon.stub(window, 'fetch');
+  const body = { "score": SCORE, "dataAvailable": true };
+  const response = { json: () => { return body }, status: 200 };
+  fetchStub.returns(Promise.resolve(response));
+
+  const done = assert.async()
+  analyzeVideo();
+  setTimeout(() => {
```
juanalonsocruz, line 17: using timeout seems like the test would be flaky. I believe Zoe has a solution for this.

src/main/webapp/scripts/analyze.js
juanalonsocruz: After this you can clean up your test from multiple types of http errors to just one.

src/main/webapp/scripts/analyze.js
```diff
@@ -0,0 +1,44 @@
...
+    updateSentimentContainer(error.message);
+  });
+}
+
+function updateSentimentContainer(text) {
+  const container = document.getElementById('sentiment-container');
+  container.innerText = text.toString();
+}
+
+function InvalidVideoException(message, metadata) {
```
chenzizhao, line 34: Is metadata used?

src/main/webapp/scripts/tests/analyze-test.js
```diff
@@ -0,0 +1,67 @@
...
+const SCORE = 0.05;
+
+QUnit.module('Updating DOM after analyzing video');
+QUnit.testStart(() => {
+  document.getElementById('video-url').innerText = VIDEO_LINK;
+})
+
+QUnit.test('DOM updates with sentiment score', function (assert) {
+  const fetchStub = sinon.stub(window, 'fetch');
+  const body = { "score": SCORE, "dataAvailable": true };
```
chenzizhao, line 11: nit: single quote

src/main/webapp/scripts/tests/analyze-test.js
```diff
@@ -0,0 +1,67 @@
+const VIDEO_LINK = "https://www.youtube.com/watch?v=testvideoid";
+const SCORE = 0.05;
+
+QUnit.module('Updating DOM after analyzing video');
+QUnit.testStart(() => {
```
chenzizhao, line 5: nit: Since you are stubbing fetch in every test and restore it afterwards, you can also extract that behavior into beforeEach and afterEach just like you did here with testStart.

src/main/webapp/scripts/analyze.js
```diff
@@ -0,0 +1,44 @@
...
+  })
+  .catch(error => {
+    updateSentimentContainer(error.message);
+  });
+}
+
+function updateSentimentContainer(text) {
+  const container = document.getElementById('sentiment-container');
+  container.innerText = text.toString();
+}
```
nif33, line 32: There's a similar helper function in script.js called updateDom, could you use that function instead? That way this code isn't repeated and it fits the MVC model better

src/main/webapp/scripts/analyze.js
kayefung: if the servlet calls ```response.sendError()``` it logs it into the console already!

src/main/webapp/scripts/analyze.js
kayefung: good point -> no it's not but I removed these exceptions because of a different comment Juan made so not a concern anymore :)

chenzizhao: gotcha

src/main/webapp/scripts/analyze.js
kayefung: nice, thanks for pointing this out!

kayefung:
The three unit tests in this PR now fail half of the time. I've figured out that it's probably caused by the call to ```updateDom()``` in ```script.js``` (it doesn't seem to actually call the method), but adding the same code directly to ```analyze.js``` works fine. 
If anyone has any ideas about why this is happening and how it can get fixed please let me know!

src/main/webapp/scripts/analyze.js
juanalonsocruz: Whoa. Didn't know that.

src/main/webapp/scripts/analyze.js
```diff
@@ -1,44 +1,23 @@
...
+async function analyzeVideo() {
   const videoId = document.getElementById('video-url').value.split('v=')[1];
-  const baseUrl = window.location.origin;
-  let url = new URL('/sentiment', baseUrl);
-  url.searchParams.append('video-id', videoId);
 
-  fetch(url)
+  return fetch(`/sentiment?video-id=${videoId}`)
   .then(response => {
     if (response.status >= 200 && response.status <= 299) {
```
juanalonsocruz, line 6: I believe this can now be response `!response.ok()`

src/main/webapp/scripts/analyze.js
```diff
@@ -1,44 +1,23 @@
...
     }
   })
   .then(videoAnalysis => {
-    if (videoAnalysis.dataAvailable) {
-      updateSentimentContainer(videoAnalysis.score);
+    if (videoAnalysis.scoreAvailable) {
+      updateDom(videoAnalysis.score.toString(), 'sentiment-container');
     } else {
-      updateSentimentContainer("We couldn't analyze this video! Try again.")
+      updateDom('We couldn\'t analyze this video! Try again.',
```
juanalonsocruz, line 16: Telling the user to try again is letting them think that the issue will be fixed if they try again. In this case it won't. A better message would be something like "Video has no available captions or comments to analyze."

src/main/webapp/tests.html
```diff
@@ -18,5 +19,33 @@
   <script src="scripts/tests/perspective-test.js"></script>
   <script src="https://apis.google.com/js/api.js"></script>
   <script defer type="module" src="scripts/tests/comment-post-test.js"></script>
+  <script src="scripts/analyze.js"></script>
+  <script defer type="module" src="scripts/tests/analyze-test.js"></script>
```
juanalonsocruz, line 23: Not sure if this is the culprit but having all the tests in the same html might cause some issues. The scripts test could be potentially manipulating the dom at the same time. For your case, it doesn't seem like its touching the dom elements but maybe something else is happening.

kayefung: This worked! Thanks!

src/main/webapp/scripts/tests/analyze-test.js
```diff
@@ -17,8 +17,9 @@ QUnit.testStart(() => {
 });
 
 QUnit.test('DOM updates with sentiment score', function (assert) {
+  console.log('sentiment score');
```
juanalonsocruz, line 20: Nit: Remove logs in tests.

src/main/webapp/scripts/tests/analyze-test.js
```diff
@@ -17,8 +17,9 @@ QUnit.testStart(() => {
...
   const body = { 'score': SCORE, 'scoreAvailable': true };
-  const response = { json: () => { return body }, status: 200 };
+  const response = { json: () => { return body }, status: 200, ok: true };
```
juanalonsocruz, line 22: Nit-ish: I believe if you make a response object it would imitate closer to what fetch returns

https://stackoverflow.com/a/44044858
https://developer.mozilla.org/en-US/docs/Web/API/Response/Response

src/main/webapp/tests.html
```diff
@@ -15,7 +15,6 @@
     <div id="menu"></menu>
   </div>
   <script src="https://code.jquery.com/qunit/qunit-2.10.0.js"></script>
-  <script src="https://unpkg.com/qunit-dom/dist/qunit-dom.js"></script>
```
juanalonsocruz, line 18: Keep this line?

kayefung: This line shows up twice, I was deleting one of the copies. Should have left a comment about it, my bad!

src/main/webapp/scripts/tests/analyze-test.js
```diff
@@ -0,0 +1,67 @@
...
+});
+
+QUnit.testStart(() => {
+  document.getElementById('video-url').innerText = VIDEO_LINK;
+});
+
+QUnit.test('DOM updates with sentiment score', function (assert) {
+  const body = { score: SCORE, scoreAvailable: true };
+  const blob = new Blob([JSON.stringify(body, null, 2)], {type : 'application/json'});
+  const init = { "status" : 200 , "statusText" : "success" };
```
chenzizhao, line 22: nit: single quote lol

kayefung: I've been writing too much java loool

PR #21, opened 2020-07-22
Fix bug with creating LanguageServiceClient instances.

kayefung:
Found a problem where you could not run ```doGet()``` in SentimentServlet and calculate the sentiment score more than once without shutting down the server and starting it up again. Realized it was caused by SentimentServlet having one instance of LanguageServiceClient that gets closed after ```calculateSentimentScore()``` is called the first time. 

Made changes to SentimentServlet and updated tests accordingly.

### DIFF
--- a/src/main/java/com/google/sps/servlets/SentimentServlet.java
+++ b/src/main/java/com/google/sps/servlets/SentimentServlet.java
@@ -93,20 +93,14 @@ public class SentimentServlet extends HttpServlet {
   /**
    * Calculates sentiment score of text. The score is from -1 (negative) to +1 (positive).
    */ 
-  private Float calculateSentimentScore(String text) throws ApiException, IOException {
-    LanguageServiceClient languageService = createLanguageServiceClient();
+  private Float calculateSentimentScore(String text) throws ApiException {
     Document doc = Document.newBuilder()
         .setContent(text)
         .setTypeValue(Document.Type.PLAIN_TEXT_VALUE)
         .build();
     sentiment = languageService.analyzeSentiment(doc).getDocumentSentiment();
     Float score = new Float(sentiment.getScore());
-    languageService.close();
 
     return score;
-  }
-
-  public LanguageServiceClient createLanguageServiceClient() throws IOException {
-    return LanguageServiceClient.create();
   }
 }
--- a/src/main/java/com/google/sps/servlets/SentimentServlet.java
+++ b/src/main/java/com/google/sps/servlets/SentimentServlet.java
@@ -83,7 +83,8 @@ public class SentimentServlet extends HttpServlet {
   /**
    * Calculates sentiment score of text. The score is from -1 (negative) to +1 (positive).
    */ 
-  private Float calculateSentimentScore(String text) {
+  private Float calculateSentimentScore(String text) throws ApiException, IOException {
+    LanguageServiceClient languageService = createLanguageServiceClient();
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -94,4 +95,8 @@ private Float calculateSentimentScore(String text) {
 
     return score;
   }
+
+  public LanguageServiceClient createLanguageServiceClient() throws IOException {
+    return LanguageServiceClient.create();
```
ajpaon, line 100: Any strong reason to make this a separate method instead of calling LSC.create directly in calculateSentimentScore?

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -83,7 +83,8 @@ private String requestToString(HttpServletRequest request) throws IOException {
   /**
    * Calculates sentiment score of text. The score is from -1 (negative) to +1 (positive).
    */ 
-  private Float calculateSentimentScore(String text) {
+  private Float calculateSentimentScore(String text) throws ApiException, IOException {
+    LanguageServiceClient languageService = createLanguageServiceClient();
```
ajpaon, line 87: just want to make sure I understand the problem. Are we not able to reuse instances of LanguageServiceClient?

src/main/java/com/google/sps/servlets/SentimentServlet.java
kayefung: Calling LSC.create() directly in the method would replace the mock of LSC that I need for testing, this is the work around Juan and I came up with in our 1:1 yesterday :/ By using this helper method we can stub it to return the mock.

I also tried using PowerMockito to stub create() for any instance of LSC but wasn't able to get it to work. Might try it again later today.

src/main/java/com/google/sps/servlets/SentimentServlet.java
kayefung: Yeah if you try and reuse it you'll run into a RejectedExecutionException

ajpaon: After reading the documentation I think the problem is not that we can't call languageService.analyzeSentiment more than once per instance. I think the problem is that we shouldn't call languageService.close until we're done with the instance.

The original implementation calls LanguageService.create in the constructor, then calls languageService.close (line 94 in this revision) in the first call to calculateSentimentScore. The documentation [1] says that future behavior is undefined. So in the next call to calculateSentimentScore we can't expect reasonable behavior from languageService.

Instead of creating a new client for each request, can we simply delete the call to close?

[1] https://googleapis.dev/nodejs/language/latest/v1.LanguageServiceClient.html#close

juanalonsocruz: @ajpaon I was concerned about leaving an open connection indefinitely (while the server is running). I feel like LanguageService API might force close the connection after some timeout.

ajpaon: It's fine to keep a grpc channel open indefinitely. Here's a thread on the grpc-java repo that discusses this: https://github.com/grpc/grpc-java/issues/3268

kayefung:
Andrew's suggestion ended up working! Went with that because I felt like it was more readable.

PR #22, opened 2020-07-22
Accept Builder as the single VideoAnalysis constructor parameter

ajpaon:
This allows us to do all assignment to VideoAnalysis in the
constructor. So we can mark all our fields final and remove setters on
VideoAnalysis. It also allows us to avoid some assignments in the
VideoAnalysis constructor that always get overwritten during build.

reference: Effective Java item #2

### DIFF
--- a/src/main/java/com/google/sps/servlets/SentimentServlet.java
+++ b/src/main/java/com/google/sps/servlets/SentimentServlet.java
@@ -57,7 +57,7 @@ public class SentimentServlet extends HttpServlet {
     String captions = captionService.getCaptionFromId(videoId);
 
     if (comments.isEmpty() && captions.isEmpty()) {
-      VideoAnalysis videoAnalysis = new VideoAnalysis.Builder()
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -57,7 +57,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     String captions = captionService.getCaptionFromId(videoId);
 
     if (comments.isEmpty() && captions.isEmpty()) {
-      VideoAnalysis videoAnalysis = new VideoAnalysis.Builder()
+      VideoAnalysis videoAnalysis = VideoAnalysis.builder()
```
kayefung, line 60: There's one more call to Builder that'd need to get updated, it's at the end of the doGet method

ajpaon: Done, thanks. Not sure why I didn't get a build error from mvn for this.

PR #25, opened 2020-07-24
Integrate UI features and Post Comment features with homepage

nif33:
-change post comment to use video id input from page
-adds bootstrap and material ui font to index.html
-adds header element and menu element
-readds a stylesheet attribute for link colour

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -1,51 +1,52 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  <title>Capstone Project</title>
+  <title>YouTube Vibe Check</title>
   <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css" integrity="sha384-9aIt2nRpC12Uk9gS9baDl411NQApFmC26EwAOH8WgZl5MYYxFfc+NcPb1dKGj7Sk" crossorigin="anonymous">
+  <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
   <link rel="stylesheet" href="style.css">
 </head>
-<body>
-  <div id="content">
+<body onload="createMenu(); createHeader()">
+  <header></header>
+  <div class="row no-gutters">
+    <ul id="menu" class="nav flex-column"></ul>
+    <div id="content">
+      <form>
+        <label for="video-url">Video URL:</label>
+        <textarea name="video-url" id="video-url"></textarea>
+        <br/><br/>
 
-    <form>
-      <label for="video-url">Video URL:</label>
-      <textarea name="video-url" id="video-url"></textarea>
-      <br/><br/>
+        <input type="button" value="Analyze" onclick="analyzeVideo()"/>
+      </form>
+      <p id="sentiment-container"></p>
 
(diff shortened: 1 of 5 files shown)


src/main/webapp/tests.html
```diff
@@ -13,6 +13,7 @@
   <div id="qunit-fixture">
     <header></header>
     <div id="menu"></div>
+    <textarea id="video-url" value="gjKpAAezWQA"></textarea>
```
chenzizhao, line 16: Is it video id?

src/main/webapp/index.html
```diff
@@ -1,51 +1,52 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  <title>Capstone Project</title>
+  <title>YouTube Vibe Check</title>
   <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
```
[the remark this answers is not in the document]
nif33: I searched for information about the [meta tag ](https://www.w3schools.com/tags/tag_meta.asp) and setting viewports and found that although it is a css-altering tag, it [should not be accessed in the CSS file](https://stackoverflow.com/questions/42515881/can-html-meta-tag-be-modified-in-css-file)

src/main/webapp/index.html
```diff
@@ -1,51 +1,52 @@
...
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css" integrity="sha384-9aIt2nRpC12Uk9gS9baDl411NQApFmC26EwAOH8WgZl5MYYxFfc+NcPb1dKGj7Sk" crossorigin="anonymous">
```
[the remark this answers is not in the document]
nif33: I got this from the bootstrap documentation. Upon further inspection, it seems a new integrity string is generated every time that [page](https://getbootstrap.com/docs/4.0/getting-started/introduction/) is accessed from a new browser

src/main/webapp/tests.html
nif33: changed the value to a url to be consistent

src/main/webapp/index.html
```diff
@@ -1,51 +1,52 @@
...
-      <br/><br/>
-
-      <input type="button" value="Is this Toxic?" onclick="getToxicity()"/>
-    </form>
-
-    <p id="toxicity-container"></p>
-
+<body onload="createMenu(); createHeader()">
+  <header></header>
+  <div class="row no-gutters">
```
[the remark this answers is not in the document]
nif33: This is the only way I saw in the documentation. From what I read, I think Bootstrap's goal is to help users avoid CSS as much as possible 😆 so they have a set of preformatted css class tags (such as the row and col) which can be referenced by classes, and overwritten by the user in their own css if needed. Eg. I overwrite parts of the nav-item tag like font-size but the Bootstrap class provides the vertical column and scaling CSS

src/main/webapp/style.css
```diff
@@ -36,6 +36,9 @@
 a {
   color: black;
 }
+a:hover {
+  color: black;
+}
```
kayefung, line 41: Nit: I feel like I've seen Zoe comment on this before, but is this necessary? Doesn't color stay the same on hover by default?

nif33: I changed it in that PR but then I realized that it was necessary after all, links by default are blue, and the hover by default is a different(?) blue, but overwriting the link colour does not overwrite the hover colour, so I'm adding it back.

kayefung: Got it, thanks for clarifying

PR #26, opened 2020-07-27
Calculate sentiment for individual comments.

kayefung:
This PR updates the implementation of the SentimentServlet class so that the sentiment scores of captions and individual comments are calculated asynchronously using Java Future and then average for a final score.


src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +56,62 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      String json = new Gson().toJson(videoAnalysis);
-
+    if (commentsList.isEmpty() && captions.isEmpty()) {
+      String json = createResponseJson(null, false);
       response.setContentType("application/json;");
       response.getWriter().println(json);
       return;
     }
+    
+    ExecutorService executor = Executors.newFixedThreadPool(26);
```
chenzizhao, line 69: why is the size 26?

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +56,62 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    Float score = null;
+    Future<Float> captionFuture = executor.submit(new Callable<Float>() {
+      @Override
+      public Float call() {
+        return calculateSentimentScore(captions);
+      }
+    });
+    
+    Float commentsScore = 0f;
+    Float captionsScore = null;
```
chenzizhao, line 90: Are they intentionally initialized in different ways?

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +56,62 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    VideoAnalysis videoAnalysis = new VideoAnalysis.Builder()
-        .setScore(score)
-        .setScoreAvailable(true)
-        .build();
-    String json = new Gson().toJson(videoAnalysis);
+    executor.shutdown();
 
+    String json = "";
+    if (!commentsList.isEmpty() && !captions.isEmpty()) {
+      json = createResponseJson((commentsScore + captionsScore) / 2f, true);
```
chenzizhao, line 108: Maybe add some comments for this class on how the scores are calculated, so that they are easier to find?

src/main/java/com/google/sps/servlets/SentimentServlet.java
kayefung: 25 comments + 1 caption, I can add a comment to specify it

src/main/java/com/google/sps/servlets/SentimentServlet.java
kayefung: hmm I guess they could both be initialized to 0, but basically I had null for both at first and then realized commentsScore has to be 0 because you're incrementing it as you retrieve the score from each comment. I'll change it so that they're both 0

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +63,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       response.setContentType("application/json;");
       response.getWriter().println(json);
       return;
     }
+    
+    // Creates 26 thread pools (25 for comments, 1 for captions)
+    ExecutorService executor = Executors.newFixedThreadPool(26);
 
-    String text = comments + ". " + captions;
+    List<Future> commentScoreFutures = new ArrayList<>();
```
ajpaon, line 79: List<Future\<Float\>>, right?

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +63,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    for (String comment : commentsList) {
+      Future<Float> future = executor.submit(new Callable<Float>() {
+        @Override
+        public Float call() {
+          return calculateSentimentScore(comment);
+        }
+      });
+      commentScoreFutures.add(future);
+    }
```
ajpaon, line 88: You can simplify this to a single line using a method reference and stream.map. Something like:

```
List<Future<Float>> commentScores = commentsList.stream().map(this::calculateSentimentScore).collect(Collectors.toList());
```

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -29,15 +35,20 @@
...
-  private final String READER_ERROR = "Reading video ID failed.";
 
   public SentimentServlet() throws IOException, GeneralSecurityException {
     languageService = LanguageServiceClient.create();
     captionService = new Caption();
     commentService = new CommentService();
   }
 
+  /**
+   * Retrieves the comments and captions of a YouTube video, given its video ID, and calculates the
```
ajpaon, line 46: this method seems to have a lot of responsibilities now. I suggest breaking some of those responsibilities (like calling the sentiment API) out into separate classes. That should make each component easier to understand and test

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +63,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      }
+    });
+    
+    Float commentsScore = 0f;
+    Float captionsScore = 0f;
 
     try {
-      score = calculateSentimentScore(text);
-    } catch (ApiException e) {
+      for (Future<Float> future : commentScoreFutures) {
```
ajpaon, line 101: you can probably simplify this using [Stream::average](https://docs.oracle.com/javase/8/docs/api/java/util/stream/IntStream.html#average--)

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +63,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     }
 
-    VideoAnalysis videoAnalysis = new VideoAnalysis.Builder()
-        .setScore(score)
-        .setScoreAvailable(true)
-        .build();
-    String json = new Gson().toJson(videoAnalysis);
+    executor.shutdown();
 
+    String json = "";
```
ajpaon, line 114: whenever I see a pattern like this:

```
String x;
if (..) {
  // assigns x
} else {
  // assigns x
}
```
I tend to pull out a helper function instead

```
String x = determineX();
// ...
String determineX() {
  if (...) {
    // returns x
  } else {
    // returns x
  }
}
```

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -103,4 +137,15 @@ private Float calculateSentimentScore(String text) throws ApiException {
 
     return score;
   }
+
+  /**
+   * Creates a Json String of a VideoAnalysis object.
+   */ 
+  private String createResponseJson(Float score, boolean available) {
```
ajpaon, line 144: I suggest trying to use primitive types (e.g. float instead of Float) wherever possible because they are not nullable.

src/main/java/com/google/sps/servlets/SentimentServlet.java
ajpaon: btw, we can probably do the same thing with commentsScore and captionsScore above

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +63,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
 
-    if (comments.isEmpty() && captions.isEmpty()) {
-      VideoAnalysis videoAnalysis = new VideoAnalysis.Builder()
-          .setScore(null)
-          .setScoreAvailable(false)
-          .build();
-      String json = new Gson().toJson(videoAnalysis);
-
+    if (commentsList.isEmpty() && captions.isEmpty()) {
+      String json = createResponseJson(null, false);
```
juanalonsocruz, line 70: I recommend you go back explicitly building the response here. The reason why it was made to a builder was to make it easy to read what was happening. Now this helper method makes it hard to read. Not obvious to know what null and false are being assigned to without reading the helper method.

I'm guessing you did this because you called this helper method in different branches below. Instead you could have branches to calculate the total average into one variable and after the branches you can build the response once.

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +63,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       response.setContentType("application/json;");
       response.getWriter().println(json);
       return;
     }
+    
+    // Creates 26 thread pools (25 for comments, 1 for captions)
+    ExecutorService executor = Executors.newFixedThreadPool(26);
```
juanalonsocruz, line 77: Make `26` into a static constant int in the class and have the comment over there.

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -52,40 +63,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      commentScoreFutures.add(future);
+    }
 
-    Float score = null;
+    Future<Float> captionFuture = executor.submit(new Callable<Float>() {
+      @Override
+      public Float call() {
+        return calculateSentimentScore(captions);
+      }
+    });
```
juanalonsocruz, line 95: If not already there, make sure to add unit tests that test that multiple comments being analyzed and that the correct average is calculated.

kayefung: Yep, the tests already account for the changes in this PR

src/main/java/com/google/sps/servlets/SentimentServlet.java
kayefung: Cool, I've never used Java Streams before! I couldn't really figure out how to use ```this::calculateSentimentScore``` because I had to use Callable. Ended up using a lambda expression ```comment -> executor.submit(callableFactory(comment))``` and made a callableFactory method that calls calculateSentimentScore from there

ajpaon: Hm, I thought you should be able to pass a lambda to a method that expects a Callable. and you should be able to use a method reference anywhere you can pass a lambda. But it's been a while since I've written java so I might be misremembering

src/main/java/com/google/sps/servlets/SentimentServlet.java
kayefung: nvm decided to modify the tests so that different strings return different sentiment scores

src/main/java/com/google/sps/servlets/SentimentServlet.java
kayefung: ack, decided not to do it because handling exceptions in a stream required a lot just to calculate an average :/

src/main/java/com/google/sps/servlets/SentimentServlet.java
```diff
@@ -111,40 +99,42 @@ public Float call() {
...
+   * to analyze (comments, captions, or both). Then returns a Json String of VideoAnalysis.
+   */
+  private String determineResponseJson(List<String> commentsList, String captions,
+      float commentsScore, float captionsScore) {
+    float score = 0f;
+    boolean scoreAvailable = true;
+
+    if (!commentsList.isEmpty() && !captions.isEmpty()) {
+      // Average commentsScore and captionsScore
+      score = new Float((commentsScore + captionsScore) / 2f);
```
juanalonsocruz, line 128: No need for `new Float`

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -34,23 +35,26 @@
...
-  @Mock(answer = Answers.RETURNS_DEEP_STUBS) Sentiment sentimentMock;
 
   @Spy HttpServletResponse responseSpy;
 
   @InjectMocks SentimentServlet sentimentServlet;
 
   @Before
-  public void setUp() throws IOException, Exception {
+  public void setUp() throws IOException {
+    when(requestMock.getParameter("video-id")).thenReturn(VIDEO_ID);
```
juanalonsocruz, line 56: Even though this is happening in every test, I believe this mock should be in each test for two reasons:

(1) Its the main input to the servlet you are testing. Each test should know what the input is for the reader.
(2) Your test mock other service requests that depend on what the input to the servlet is. For example, when you mock getCommentsFromId and getCaptionFromId you give the video id. Not obvious to the reader why that video id is or where it came from. You want the test to be read like "If the input to the servlet is VIDEO_ID and when getCommentsFromId gets VIDEO_ID then return blah and in the end the behavior blah."

src/test/java/com/google/sps/SentimentServletTest.java
```diff
@@ -65,16 +69,17 @@ public void printsVideoAnalysisJsonObject() throws IOException {
     List<String> comments = Arrays.asList("foo","bar","foobar");
     String captions = "foo. bar. foobar.";
 
-    when(requestMock.getParameter("video-id")).thenReturn(VIDEO_ID);
     when(commentServiceMock.getCommentsFromId(VIDEO_ID)).thenReturn(comments);
     when(captionServiceMock.getCaptionFromId(VIDEO_ID)).thenReturn(captions);
-    when(languageServiceMock.analyzeSentiment(any(Document.class)).getDocumentSentiment())
-        .thenReturn(sentimentMock);
-    when(sentimentMock.getScore()).thenReturn(SCORE);
+    when(sentimentAnalysisMock.calculateSentimentScore(comments.get(0))).thenReturn(SCORE_1);
```
juanalonsocruz, line 74: Super extreme nit: Index numbers don't match with score "index".

PR #27, opened 2020-07-27
Remove unused gapi fakes in tests

chenzizhao:
The fake gapi objects that weren't used in stubbing gapi are removed. 

They were initially created to fix the 'Cannot stub non-existent own property init' when mocking `gapi.client.init` but now it's not a problem anymore. I haven't figured out exactly why but very likely I made some mistakes using sinon in the first attempt.

### DIFF
--- a/src/main/webapp/scripts/tests/comment-post-test.js
+++ b/src/main/webapp/scripts/tests/comment-post-test.js
@@ -1,11 +1,11 @@
 import { postVideoComment } from '/scripts/post-comment.js';
+import 'https://apis.google.com/js/api.js';
 
 const VIDEO_ID = 'sample video id';
 const COMMENT_TEXT = 'sample comment';
 
 QUnit.module('Post a comment via YouTube API', {
     beforeEach: () => {
-        gapi = new CustomGapi();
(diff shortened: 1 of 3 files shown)


src/main/webapp/scripts/tests/comment-post-test.js
```diff
@@ -5,7 +5,6 @@ const COMMENT_TEXT = 'sample comment';
 
 QUnit.module('Post a comment via YouTube API', {
     beforeEach: () => {
-        gapi = new CustomGapi();
         sinon.stub(gapi.client, 'request');
```
juanalonsocruz, line 8: What scope is this gapi from? Is it because the DOM included the other js file that contains it? I'm going to approve this PR but I find it odd that the reader kind of has to guess where this variable is coming from. It would be easier to read to if it was explicitly imported.

chenzizhao: Done. This gapi was from `<script src="https://apis.google.com/js/api.js"></script>` in tests.html, but I just moved it to the respective test js files for readability.

PR #28, opened 2020-07-27
Clear commentButtonContainer before appending button

chenzizhao:
The tests revealed the bug that when the user signs in and out multiple times, there will be multiple comment buttons displayed with the same id. So the fix is to clear the container before appending the button.


chenzizhao:
This case was not tested in the original PR. The bug was later caught by @nif33. I will add a new test case to include this case.

PR #29, opened 2020-07-28
Fix NullPointerException in parsing caption

chenzizhao:
@kayefung points out that Caption.getCaptionFromId() throws uncaught NullPointerException for some videos (https://www.youtube.com/watch?v=X-euQU-WcJY for example) caused by this line:

```
caption.append(textDom.item(i).getFirstChild().getNodeValue());
```

This is because some text dom items are empty, so `.getFirstChild()` returns null.

```
<text start="133.07" dur="1.2">WOMAN 4: I&#39;m joking! I&#39;m joking!</text>
<text start="134.27" dur="1.22"/> // This line will cause the NullPointerException
``` 

The current fix is to wrap it with try-catch block.

### DIFF
--- a/src/main/java/com/google/sps/servlets/Caption.java
+++ b/src/main/java/com/google/sps/servlets/Caption.java
@@ -1,10 +1,12 @@
 package com.google.sps.servlets;
 
+import java.io.Console;
 import java.io.IOException;
 import java.io.InputStream;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import org.w3c.dom.Document;
+import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
@@ -38,8 +40,11 @@ public class Caption {
     NodeList textDom = doc.getElementsByTagName("text");
     StringBuilder caption = new StringBuilder();
     for (int i = 0; i < textDom.getLength(); i++) {
-      caption.append(textDom.item(i).getFirstChild().getNodeValue());
-      caption.append(" ");
+      Node child = textDom.item(i).getFirstChild();
+      if (child != null) {
+        caption.append(textDom.item(i).getFirstChild().getNodeValue());
+        caption.append(" ");
+      }
     }
     return caption.toString().trim();
   }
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -38,8 +38,13 @@ private String parseXmlFromStream(InputStream inputStream)
     NodeList textDom = doc.getElementsByTagName("text");
     StringBuilder caption = new StringBuilder();
     for (int i = 0; i < textDom.getLength(); i++) {
-      caption.append(textDom.item(i).getFirstChild().getNodeValue());
-      caption.append(" ");
+      try {
+        caption.append(textDom.item(i).getFirstChild().getNodeValue());
+        caption.append(" ");
+      } catch (NullPointerException ok) {
+        // getFirstChild() returns null if the item has no text
```
ajpaon, line 45: instead of catching NPE you could consider checking for null before calling getNodeValue. something like:

```
const child = textDom.item(i).getFirstChild();
if (child) {
  caption.append(child.getNodeValue());
  caption.append(" ");
}
```

I like that approach because I feel like it removes the need for the comment in your catch block.

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -38,8 +38,13 @@ private String parseXmlFromStream(InputStream inputStream)
...
+        continue;
+      }
```
juanalonsocruz, line 47: Add a unit test for this case.

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -38,8 +38,13 @@ private String parseXmlFromStream(InputStream inputStream)
     NodeList textDom = doc.getElementsByTagName("text");
     StringBuilder caption = new StringBuilder();
     for (int i = 0; i < textDom.getLength(); i++) {
-      caption.append(textDom.item(i).getFirstChild().getNodeValue());
-      caption.append(" ");
+      try {
+        caption.append(textDom.item(i).getFirstChild().getNodeValue());
```
juanalonsocruz, line 42: Why can it be null in the first place?

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Done. Thanks for the suggestion!

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Done

src/main/java/com/google/sps/servlets/Caption.java
chenzizhao: Not entirely sure how captions work, it seems like each line of caption has max duration and it is an empty item if no one speaks during some period of time.

PR #30, opened 2020-07-29
Add user instructions and more UI to homepage

nif33:
small changes to improve user experience:
-index.html: use bootstrap forms to display input areas and add text instructions
-analyze.js and script.js: add functions that display sentiment score in a pretty and colourful fraction with a fun loading spinner
-perspective.js: make string 1 decimal, and add explanation of what the number means
-view.js: change comment buttons to inputs, and change text to log in, log out, post
-script.js: add a logo to the header
Screenshot 2020-07-28 at 6 01 16 PM - Display 2


src/main/webapp/scripts/analyze.js
```diff
@@ -10,14 +10,10 @@ async function analyzeVideo() {
...
-    } else {
-      updateDom('Video has no available captions or comments to analyze.',
-          'sentiment-container');
-    }
+    displaySentiment(videoAnalysis);
   })
   .catch(error => {
-    updateDom(error.message, 'sentiment-container');
+    displaySentiment({ scoreAvailable: false });
+    console.error(error);
```
nif33, line 17: changed this to log the error in the console and then use the logic in displaySentiment to show an error to the user

src/main/webapp/scripts/view.js
```diff
@@ -78,7 +81,7 @@ function successfulApiCallback(data) {
  */
 function failedApiCallback(err) {
     console.error('execution error ', err);
-    updateDom(err, postResultContainer);
+    updateDom('An error occurred for posting this comment.', postResultContainer);
```
nif33, line 84: changed this because it was giving the user long and confusing errors meant for the console when they did things like not having a video to comment on

src/main/webapp/scripts/analyze.js
```diff
@@ -1,6 +1,6 @@
 async function analyzeVideo() {
   const videoId = document.getElementById('video-url').value.split('v=')[1];
```
juanalonsocruz, line 2: Nit: Could this be changed into a regex to capture the videoId? Users might be in some weird version of the URL with more stuff after the video id.
https://stackoverflow.com/a/8260383

src/main/webapp/scripts/view.js
chenzizhao: Actually the error gives some useful information like comments can't be empty, or the video isn't found etc. I think it is informative to show the user that piece?

nif33: This is the one that bothered me, happens when a user didnt input a video. the first sentence is super helpful but the rest is ??? not even sure why there's all those tags, maybe something broke with the API because that's what console logs too. Maybe in the future it won't be a giant code block or we can write a function that converts errors like these into nice text that a user can read.
Screenshot 2020-07-29 at 3 09 09 PM

src/main/webapp/scripts/analyze.js
nif33: Changed. Wow there were URL's I haven't even heard of before, what in the world is youtube-nocookie?

PR #31, opened 2020-07-29
Extension: add context menu items and notifications

chenzizhao:
This chrome extension allows the user to check the video vibe and comment toxicity with Chrome's own context menu and notification. It reduces the friction to use our service because the user no longer needs to copy the link or text, open our website and then analyze. 

The chrome extension will be connected to our backend in a later PR. Currently the handler always returns the same string. 

Thank you @nif33 for setting up the chrome extension files and sharing your notes.

### DIFF
--- a/chrome-extension/background.js
+++ b/chrome-extension/background.js
@@ -1,12 +1,89 @@
-chrome.runtime.onInstalled.addListener(function() {
-  chrome.declarativeContent.onPageChanged.removeRules(undefined, function() { // undefined removes all rules
+const MS_BEFORE_CLEAR_NOTIFICAITON = 2000;
+const VIBE_CHECK_MENU_ID = 'vibeCheck';
+const TOXICITY_CHECK_MENU_ID = 'toxicityCheck';
+const ICON_PATH = '/images/temp-icon.png';
+
+chrome.runtime.onInstalled.addListener(function () {
+  // "undefined" removes all rules
+  chrome.declarativeContent.onPageChanged.removeRules(undefined, function () {
     chrome.declarativeContent.onPageChanged.addRules([{
       // required conditions for the extension to be active
       conditions: [new chrome.declarativeContent.PageStateMatcher({
-        pageUrl: {hostEquals: 'www.youtube.com'},
+        pageUrl: { hostEquals: 'www.youtube.com' },
       })],
       // things to do if the extension is active
       actions: [new chrome.declarativeContent.ShowPageAction()]
     }]);
   });
 });
+
+chrome.runtime.onInstalled.addListener(function () {
+  const createProperties = {
+    'id': VIBE_CHECK_MENU_ID,
+    'title': 'Vibe check this video',
+    'contexts': ['link'],
+  };
+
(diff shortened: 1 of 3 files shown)


chrome-extension/background.js
```diff
@@ -1,12 +1,83 @@
...
+  };
+
+  chrome.contextMenus.create(createProperties);
+});
+
+chrome.contextMenus.onClicked.addListener(async function (info) {
+  if (info.menuItemId == VIBE_CHECK_MENU_ID) {
+    const notificationOption = {
+      type: 'basic',
+      iconUrl: '/images/temp-icon.png',
```
juanalonsocruz, line 42: Nit: Put this string into a constant above so it could be changed in the future easily

chrome-extension/background.js
```diff
@@ -1,12 +1,83 @@
...
+  const createProperties = {
+    'id': TOXICITY_CHECK_MENU_ID,
+    'title': 'Toxicity check this line',
+    'contexts': ['selection'],
+  };
+
+  chrome.contextMenus.create(createProperties);
+});
+
+chrome.contextMenus.onClicked.addListener(async function (info) {
```
juanalonsocruz, line 38: Is it common practice to have multiple listeners on the same object and with same trigger? (in this case when contextMenus is clicked). Is there something technically bad about it? If there isn't, I would just have one listener with a switch statement on menuItemId and then a helper method to handle that case.

chrome-extension/background.js
```diff
@@ -1,12 +1,83 @@
-chrome.runtime.onInstalled.addListener(function() {
-  chrome.declarativeContent.onPageChanged.removeRules(undefined, function() { // undefined removes all rules
+const MS_BEFORE_CLEAR_NOTIFICAITON = 2000;
+const VIBE_CHECK_MENU_ID = 'vibe check';
+const TOXICITY_CHECK_MENU_ID = 'toxicity check';
+
+chrome.runtime.onInstalled.addListener(function () {
+  chrome.declarativeContent.onPageChanged.removeRules(undefined, function () { // undefined removes all rules
```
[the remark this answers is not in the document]
chenzizhao: done

chrome-extension/background.js
chenzizhao: done

chrome-extension/background.js
```diff
@@ -1,12 +1,83 @@
-chrome.runtime.onInstalled.addListener(function() {
-  chrome.declarativeContent.onPageChanged.removeRules(undefined, function() { // undefined removes all rules
+const MS_BEFORE_CLEAR_NOTIFICAITON = 2000;
+const VIBE_CHECK_MENU_ID = 'vibe check';
+const TOXICITY_CHECK_MENU_ID = 'toxicity check';
```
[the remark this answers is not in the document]
chenzizhao: Well it doesn't look right lol. Changed to camel case

chrome-extension/background.js
chenzizhao: Not sure about common practice but a switch statement is always nicer here!

PR #33, opened 2020-07-29
Extension: connect to sentiment analysis service

chenzizhao:
I extracted the functions that calls the server (`getToxicityAnaylsisMsg`, `getSentimentAnaylsisMsg`) from background.js. The new file background.html allows for importing in background.js. 

I implemented `getSentimentAnaylsisMsg()` the same as `analyzeVideo()` in src/main/webapp/scripts/analyze.js. However, instead of GET from the servlet `/sentiment?video-id=`, I requested to `https://step70-2020.appspot.com/sentiment?video-id=`. 

**Issue**: Cross-origin resource sharing (CORS)
- If the request mode is set to `cors` (default), it returns an error 
```
Access to fetch at 'https://step70-2020.appspot.com/sentiment?video-id=QUmbwTD8pAY' 
from origin 'chrome-extension://ccoccngopkngafeanfogojgllljpfghd' has been blocked by CORS policy: 
No 'Access-Control-Allow-Origin' header is present on the requested resource. 
If an opaque response serves your needs, set the request's mode to 'no-cors' 
to fetch the resource with CORS disabled.
```
- If the request mode is set to `no-cors`, the correct response is received. The content can be viewed in chrome dev tools but is hidden from the front end javascript (response.status=0, response.ok=false, response.message=undefined). 
- I referenced this thread https://stackoverflow.com/questions/43262121/trying-to-use-fetch-and-pass-in-mode-no-cors and got around the problem by adding a CORS proxy. I doubt if this is a good/safe solution. Any thoughts?


ajpaon:
Instead of using a proxy, does the request work if we request permission for https://step70-2020.appspot.com/ in the manifest?

As described here: https://developer.chrome.com/extensions/xhr#requesting-permission

chenzizhao:
Thanks it worked! I should've asked sooner

chrome-extension/service.js
```diff
@@ -0,0 +1,37 @@
+/**  
+ * @param {String} linkUrl, to be analyzed
+ * @returns {String} sentiment analysis message
+ */
+async function getSentimentAnaylsisMsg(linkUrl) {
+  // TODO use regular expression 
+  const videoId = linkUrl.split('v=')[1];
```
nif33, line 7: I have made a function that gets parses url to id using regex. its in script.js called `getVideoId` I think, can you copy it and use it over here

chrome-extension/service.js
```diff
@@ -0,0 +1,37 @@
...
+    .then(response => {
+      if (response.ok) {
+        return response.json();
+      } else {
+        throw new Error(response.statusText);
+      }
+    })
+    .then(videoAnalysis => {
+      if (videoAnalysis.scoreAvailable) {
+        return videoAnalysis.score.toString() + ' out of ±1';
```
nif33, line 18: I changed this in a recent PR to parse it into a nice fraction /10 (I read that it is good to convert to scales that users are familiar with), the function to do this is sentimentScoreToNumerator(score) in script.js, could you use it here.

chrome-extension/service.js
chenzizhao: Done thanks

chrome-extension/service.js
chenzizhao: Done thanks!

PR #36, opened 2020-07-31
Display videos on popular.js page

nif33:
video-class.js adds a classes and stores the important data from the list video API call in a easy-to-access way.
recommended-video-display.js handles creating video cards from Video objects and adding them to the page.
currently, cards do not link to anything (this is changed in a later PR)
tests were separated into 3 files, and popular-test.js was modified for the new video display

Screenshot 2020-07-30 at 4 59 20 PM


src/main/webapp/comment-tests.html
```diff
@@ -0,0 +1,32 @@
...
+  <script src="https://unpkg.com/qunit-dom/dist/qunit-dom.js"></script>
+  <script src="https://sinonjs.org/releases/sinon-9.0.0.js"></script>
+  <script src="https://apis.google.com/js/api.js"></script>
+  <script src="scripts/script.js"></script>
+  <script type="module" src="scripts/tests/comment-post-test.js"></script>
+  <script defer type="module" src="scripts/tests/comment-post-test.js"></script>
+  <script defer type="module" src="scripts/tests/controller-auth-test.js"></script>
+
+  <!-- Async tests seem to have some conflicts with qunit-fixture.
+  Qunit may have restored the page before assertion happened. -->
```
juanalonsocruz, line 23: Does this mean the tests are flaky?

src/main/webapp/popular.html
```diff
@@ -3,15 +3,23 @@
...
   <link rel="stylesheet" href="style.css">
 </head>
-<body>
-  <div id="content">
-    <label for="region">Choose a region:</label>
-    <select name="region" id="region-select">
-      <option value="US">United States</option>
-    </select>
-    <div id="popular-list-container"></div>
+<body onload="createMenu(); createHeader()">
```
juanalonsocruz, line 10: Nit: Having more than one method in html seems smelly to me. I'm conflicted between keeping static html in this html instead of dynamically creating in js. It seems like these two methods create static content. I would ask Amir for an opinion on this.

src/main/webapp/scripts/classes/recommended-video-display-class.js
```diff
@@ -0,0 +1,49 @@
+class RecommendedVideoDisplay{
```
juanalonsocruz, line 1: Super Extreme Nit (no need to change at all): This class seems very re-usable for any other use case besides recommendations. It would be fine it was called VideoDisplay.

src/main/webapp/scripts/classes/recommended-video-display-class.js
```diff
@@ -0,0 +1,49 @@
...
+    const title = this.newElement(this.#TITLE);
+    const text = this.newElement(this.#TEXT);
+    thumbnail.src = video.thumbnail;
+    title.innerText = video.title;
+    title.title =  video.title;
+    if(video.sentiment === null) {
+      text.innerText = video.likeCount + ' likes';
+    } else {
+      text.innerText = video.sentiment;
+    }
```
juanalonsocruz, line 34: Nit: Kind of inconsistent behavior to have one or the other. If there is no score, then is it actually a popular video? Oh or maybe I have this wrong. Is this the trending page results? Or popular videos (based on sentiment score) that are stored in the datastore?

src/main/webapp/scripts/popular.js
```diff
@@ -1,13 +1,15 @@
...
 function loadPopular() {
-  return loadApi().then(popular.getTrendingFromYoutubeApi)
+  return popular.getTrendingFromYoutubeApi()
   .then((response) => {
+    const recommended = new RecommendedVideoDisplay('popular-list-container');
     for(const item of response.items) {
-      // Ready in a later PR: display the item nicely on the page
-      const containerDOM = document.getElementById('popular-list-container');
-      containerDOM.innerText += item;
+      const video = new Video(item, null);
```
juanalonsocruz, line 11: Seems odd that sentiment is not being added. Is it because it will be in another PR? Add a TODO/issue for this if so.

src/main/webapp/comment-tests.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
```
juanalonsocruz, line 1: General comment on UI and branding. No comment on whether you should change what you have now or not.

https://www.youtube.com/about/brand-resources/
https://developers.google.com/youtube/terms/branding-guidelines

src/main/webapp/comment-tests.html
nif33: this was Zoe's comment indicating that the html is here instead of qunit-fixture because tests were flaky otherwise, so they're not flaking anymore

src/main/webapp/scripts/popular.js
nif33: Sentiment is not used for the popular videos, but it will be used for the page that displays top-sentiment videos

src/main/webapp/scripts/classes/recommended-video-display-class.js
nif33: I did this so that the function can be reused for the positive videos page so like, if there's a sentiment (this video was from the datastore) then display it, if not (this video is from youtube API) display # of likes

src/main/webapp/popular.html
nif33: now calling one function (addMenuHeader()) that calls function to create a menu and header.
I made it dynamic because as I was editing and making changes I had to apply them to every html page, which was a pain. Curious to get @FuzzyTree 's opinion on this though, since I know this method requires that js be loaded in order for the user to navigate the website which is a downside of dynamically creating these elements

src/main/webapp/scripts/classes/recommended-video-display-class.js
nif33: I named it thinking we might need a page that just displays one video but that's not happening anymore haha. I'll call it VideoListDisplay just so that it's clear there are many videos here

PR #37, opened 2020-08-01
link videos from video list display to index.html so it can be analyzed

nif33:
clicking a trending video will now take you to index.html where that video can be analyzed
also added an iframe video player in the homepage

### DIFF
--- a/src/main/webapp/scripts/analyze.js
+++ b/src/main/webapp/scripts/analyze.js
@@ -1,5 +1,11 @@
+let player;
+function onYouTubeIframeAPIReady() {
+  player = new YT.Player('player', {});
+}
+
 async function analyzeVideo() {
   const videoId = getVideoId();
+  player.loadVideoById(videoId, 0, "large");
   displayLoading('sentiment-container');
   return fetch(`/sentiment?video-id=${videoId}`)
   .then(response => {
--- a/src/main/webapp/scripts/analyze.js
+++ b/src/main/webapp/scripts/analyze.js
@@ -1,11 +1,6 @@
-let player;
-function onYouTubeIframeAPIReady() {
-  player = new YT.Player('player', {});
-}
-
 async function analyzeVideo() {
   const videoId = getVideoId();
-  player.loadVideoById(videoId, 0, "large");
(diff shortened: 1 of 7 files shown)


src/main/webapp/scripts/script.js
```diff
@@ -121,3 +123,16 @@ function displayLoading(containerId) {
...
 }
+
+/* Checks URL params for if there is a video that needs to be searched */
+function searchVideoFromParams() {
+  queryString = window.location.search;
+  if(queryString) {
+    document.getElementById('video-url').value = 'https://www.youtube.com/watch' + window.location.search;
+    analyzeVideo();
+  }
+}
```
nif33, line 134: is this the best way to do this? No.
is it the easiest way to do it with the code we currently have and within the last week we have? Yeah :/

src/main/webapp/scripts/tests/view-test.js
```diff
@@ -7,3 +7,7 @@ QUnit.test('createMenu', function(assert) {
   createMenu();
   assert.dom('.nav-item').exists({ count: 3});
 });
+
+QUnit.test('iFrame created', function(assert) {
+  assert.dom('iframe').exists({ count: 1 });
+});
```
nif33, line 13: Should I be testing that the iframe is added? honestly not sure what to test here, because the iframe api adds it, and I could mock the API, but then there is no code to test...hmm

src/main/webapp/scripts/analyze.js
```diff
@@ -1,5 +1,6 @@
 async function analyzeVideo() {
   const videoId = getVideoId();
+  player.loadVideoById(videoId);
```
chenzizhao, line 3: Could you add a comment about 'player' and where it comes from?

src/main/webapp/scripts/script.js
juanalonsocruz: thanks for clarifying :)

src/main/webapp/scripts/tests/view-test.js
juanalonsocruz: good enough for me lol.

PR #38, opened 2020-08-04
Use threads and invokeAny to get captions faster

chenzizhao:
@nif33 suggested that using threads to fetch 6 captions of the same video is faster than doing so consecutively. It is implemented here. Time taken to access captions for one video is reduced from ~2000ms to ~400ms in my experiments. 

Two tests are unstable (run tests: `mvn clean && mvn test`). As @kayefung suggested, further specifying the setup can ensure the tests to pass. What I don't understand is why the original tests become flaky once threads are introduced. 

EDIT:
When the flaky tests fail, it is due to the `ExecutionException` thrown by invokeAny, which indicates *none* of the tasks complete successfully. However, the original set up should allow *all* tasks to finish. 

`ExecutionException` is caused by `SAXParseException@67 "org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 10; White space is required between the processing instruction target and data."` my current guess is Stream/DocumentBuilder is not threadsafe. 

RESOLVED:
It turns out the issue isn't even related to threads -- the inputstreams returned by the mocks in the tests are not reusable. The SAXParseException is thrown by every caller when all of them try to parse the same inputstream simultaneously. The fix is instead of `thenReturn`, create a new inputstream with `thenAnswer`, so every caller gets their own copy. This has been an undiscovered bug.

### DIFF
--- a/src/main/java/com/google/sps/servlets/Caption.java
+++ b/src/main/java/com/google/sps/servlets/Caption.java
@@ -1,8 +1,14 @@
 package com.google.sps.servlets;
 
-import java.io.Console;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.stream.Collectors;
+import java.util.concurrent.ExecutionException;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import org.w3c.dom.Document;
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/sps/servlets/CaptionTest.java
```diff
@@ -63,7 +63,9 @@ public void init() {
   @Test
   public void properXmlReturnsConcatenatedString() throws IOException {
     InputStream stream = new ByteArrayInputStream(PROPER_XML.getBytes());
-    when(captionServiceMock.fetchStream(eq(VIDEO_ID), contains(LANG_EN))).thenReturn(stream);
+    when(captionServiceMock.fetchStream(eq(VIDEO_ID), eq(LANG_EN))).thenReturn(stream);
```
kayefung, line 66: I might be wrong but I believe you only use eq() when other parameters use argument matchers, so I don't think it's necessary here

chenzizhao: Nice catch! it is indeed unnecessary (i replaced `contains` with `eq` from the original setup, as you may tell)

juanalonsocruz:
Love the PR message and the resolution haha.

src/main/java/com/google/sps/servlets/Caption.java
```diff
@@ -23,15 +29,26 @@
...
+        Arrays.stream(languages).map(language -> new Callable<String>() {
+          @Override
+          public String call() throws Exception {
+            return parseXmlFromStream(captionService.fetchStream(videoId, language));
+          }
+        }).collect(Collectors.toList());
+
+    String caption = "";
+    try {
+      caption = executor.invokeAny(captionCallables);
```
juanalonsocruz, line 45: Now that I think about it, does the sentiment/perspective API determine the text language is and then calculates a score?

chenzizhao: @kayefung says it the sentiment API detects the language before analyzing. Not sure about perspective API though. The "different" languages here are variants of English.

PR #39, opened 2020-08-05
add Positive Videos page and integrate datastore

nif33:
This is a lot of files but this will be my last PR, it
-creates an HTML page for positive videos (pretty much the same as popular.html)
-positive.js fetches from backend database (very similar to popular.js)
-adds videoId to VideoAnalysis class
-calls the database function to store videos after getting a sentiment score
-changes sentiment tests for new code and adds tests for frontend code


nif33:
Gonna wait for the other PR to finish because it was getting too confusing with multiple people making the same changes

